PR #12, opened 2020-04-19
mdbook version update and new guard clause docs

tomwhatmore:
Went for just a single example in the end as the float and int versions are very similar.


lpil:
Could you update this PR to target the gleam-0.8 branch please? These features are not released yet

tomwhatmore:
Is it not already?

lpil:
They're on master but we have not released a new versin from master yet.

lpil:
Ah yes! I got confused. Thank you :)

PR #14, opened 2020-04-21
Add a section for using the string module

CrowdHailer:
I think this would be helpful to add on this page.
Makes calling functions on modules show up earlier in the walkthrough.

### DIFF
--- a/src/tour/strings.md
+++ b/src/tour/strings.md
@@ -18,3 +18,15 @@ Special characters such as `"` need to be escaped with a `\` character.
 ```rust,noplaypen
 "Here is a double quote -> \" <-"
 ```
+
+## Manipulating strings.
+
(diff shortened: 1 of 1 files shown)


lpil:
This would be using the concept of imports, variables, and function calls before they are defined, so I would like to hold off on this.

Later we'll have more detailed documentation which will be more useful as an introduction to writing actual Gleam programs.

CrowdHailer:
Fair enough, though I believe that in that case it might be nice to introduce a "real" hello world example a bit earlier, just as an intro to what's coming up

lpil:
Yeah I'm envisioning having an intro which is a a step by step making a toy project, and having that first. The tour might evolve into more of a language reference and will be less prominent on the site.

PR #27, opened 2020-07-12
Add data type info to the Erlang cheatsheet

devonestes:
Hopefully this makes getting up and running a bit faster for folks
trying out Gleam who have some Erlang experience.

### DIFF
--- a/src/cheatsheets/gleam-for-erlang-users.md
+++ b/src/cheatsheets/gleam-for-erlang-users.md
@@ -11,15 +11,15 @@
 - [Operators](#operators)
 - [Constants](#constants)
 - [Blocks](#blocks) TODO
-- [Data types](#data-types) TODO
-  - [Strings](#strings) TODO
-  - [Tuples](#tuples) TODO
-  - [Lists](#lists) TODO
-  - [Atoms](#atoms) TODO
-  - [Maps](#maps) TODO
+- [Data types](#data-types)
+  - [Strings](#strings)
+  - [Tuples](#tuples)
+  - [Lists](#lists)
+  - [Atoms](#atoms)
+  - [Maps](#maps)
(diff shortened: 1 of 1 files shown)


src/cheatsheets/gleam-for-erlang-users.md
```diff
@@ -192,3 +192,163 @@ fn main() {
...
+#### Gleam
+
+```rust,noplaypen
+"Hellø, world!"
+```
+
+### Tuples
+
+Unlike in Erlang, there is no tuple literal syntax in Gleam, but they are very useful in Gleam as
+they're the only collection data type that allows for mixed types of elements in the collection.
```
lpil, line 216: This isn't quite right, we have a tuple literal syntax in Gleam! You've used it below :)

src/cheatsheets/gleam-for-erlang-users.md
```diff
@@ -192,3 +192,163 @@ fn main() {
...
+#### Gleam
+
+```rust,noplaypen
+type Person {
+  Person(age: Int, name: String)
+}
+
+let person = Person(name: "name", age: 35)
+
+let Person(name: name, age: _) = person
```
lpil, line 352: `let name = person.name` would also work

src/cheatsheets/gleam-for-erlang-users.md
devonestes: That looks a whole lot like a function call and not a literal... I guess there‘s no a reason why that can’t be a literal, but to someone just coming in to the language that would surely be mistaken as a function call!

lpil: It's a similar syntax, but it's not a function. You can't assign the `tuple` keyword to a variable, for example.

PR #31, opened 2020-08-28
Update code examples in the Language tour

michallepicki:
In the comments tour, there was an unnecessary empty line that was being rendered at the beginning of a code example.
It's now removed.

In the constants tour, code examples were missing a `noplaypen` annotation, so they had the "run" and "show hidden code" buttons showing up (not functional). It is now added.


lpil (approved):
Thank you

PR #32, opened 2020-08-29
Small grammar corrections

michallepicki:
I'm not an English native speaker, but this feels more correct-ier? :)


lpil (approved):
Yes this is correct! Your English is better than mine :)

PR #35, opened 2020-09-07
Fix typo on running the project page

Ian-GL:
Noticed that on the last paragraph of "Running the project" there is this sentence: `At a later date will will have built in support and [...]`. I guess there was a mistake and it was meant to be `we will`, or maybe `Gleam will`. This is just a fix for that.


lpil (approved):
Thanks!

PR #39, opened 2020-09-12
 Move Result type definition to a separate page in the Language Tour

michallepicki:
Because Result is a part of Prelude, I think it deserves having its own page. I simply moved the text about it that was already present on the Custom Types page.

Additionally, because the "Bool" type was already being defined on that page, I moved information about it from the "Commonly used custom types" section to a different place earlier on that page.


lpil (approved):
A good improvement, thank you

PR #44, opened 2020-10-17
Make releases links clickable

michaeljones:
And provide some text for them rather than just the urls.

Minor but hopefully useful.


lpil (approved):
Much better! Thank you

michaeljones:
Ha, you were too quick! Turns out the url for rebar needs updating too. I'll make another :)

PR #45, opened 2020-10-17
Update rebar3 release url

michaeljones:
It seems that it has moved.


lpil (approved):
oooh new website! Thanks

PR #51, opened 2020-10-28
Add gleam-for-elixir-users page

sharno:
Work in progress.
Following the structure of the page existing for Erlang. And leveraging code from Elixir's official docs.

### DIFF
--- a/cheatsheets/gleam-for-elixir-users.md
+++ b/cheatsheets/gleam-for-elixir-users.md
@@ -1,0 +1,235 @@
+---
+layout: page
+title: Gleam for Elixir users
+---
+
+- [Comments](#comments)
+- [Variables](#variables)
+  - [Match operator](#match-operator)
+  - [Variable type annotations](#variable-type-annotations)
+- [Functions](#functions)
+  - [Exporting functions](#exporting-functions)
+  - [Function type annotations](#function-type-annotations)
+  - [Function heads](#function-heads)
+  - [Function overloading](#function-overloading)
+  - [Referencing functions](#referencing-function)
+  - [Labelled arguments](#labelled-arguments)
+- [Operators](#operators)
+  - [Pipe](#pipe)
+- [Constants](#constants)
+- [Blocks](#blocks)
+- [Data types](#data-types)
+  - [Strings](#strings)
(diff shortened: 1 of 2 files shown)


cheatsheets/gleam-for-elixir-users.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### Gleam
+
+In Gleam `=` could be used for pattern matching as well, but you'll get compile errors if there's a type mismatch
+and a runtime error if there's a value mismatch. Also, there's the `assert` keyword, used to make assertions.
+
+```rust
+let [element] = "Hello" // Compile error, type mismatch
+let [element] = [1, 2] // Erlang runtime error
+let [element] = [1] // works
```
lpil, line 81: It would be preferable to use `assert` for this runtime error. In future versions of Gleam a binding that can fail will not be accepted by the compiler if `let` is used.

cheatsheets/gleam-for-elixir-users.md
```diff
@@ -0,0 +1,235 @@
...
+
+```rust
+fn sum(x, y) {
+  x + y
+}
+
+// anonymous functions are defined in a similar syntax without the name
+fn calculate(x) {
+    let add_one = fn(z) { z + 1 }
+    add_one(x)
```
lpil, line 138: Can we use 2 space indentation here please? :)

cheatsheets/gleam-for-elixir-users.md
```diff
@@ -0,0 +1,235 @@
...
+#### Gleam
+
+```rust
+fn add(x: Int, y: Int) -> Int {
+  x + y
+}
+```
+
+Unlike in Elixir these type annotations will always be checked by the compiler
+and have to be correct for compilation to succeed.
```
lpil, line 202: Could we also add something like "Type annotations are optional in Gleam and the compiler will still type check your program if they are omitted" ?

lpil (commented):
Great stuff! Thank you. I've added a couple small notes below

sharno:
@lpil I think that's good enough for this PR, it's ready for review and merge

lpil (approved):
Wonderful! Thank you!

PR #55, opened 2020-11-01
Update getting-started.md

l1x:
Adding the missing section of installing rebar3


lpil (commented):
Thank you! This only has instructions for *nix, we would also need to support Windows.

We could alternatively link to the rebar3 docs, that'll ensure it's up do date? I'm not sure it's better to have multiple sites though. What do you think?

PR #58, opened 2020-11-09
Update the Gleam for Elixir users cheatsheet

michallepicki:
I noticed there are a few mistakes in the Elixir cheatsheet (not working hyperlinks, code that doesn't compile or false statements) so I started fixing them, I noticed more inconsistencies across the document, started updating them and there went my evening. I tried to show equivalent code (where possible) between Elixir and Gleam, and not worry about things Elixir users should already know (like the fact that functions need to be defined inside modules), since the document is targeted at Elixir users learning Gleam.

### DIFF
--- a/cheatsheets/gleam-for-elixir-users.md
+++ b/cheatsheets/gleam-for-elixir-users.md
@@ -3,16 +3,15 @@ layout: page
 title: Gleam for Elixir users
 ---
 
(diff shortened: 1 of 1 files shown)


lpil (approved):
Wonderful, this is a good improvement. Thank you

PR #59, opened 2020-11-20
Add initial version of 'gleam.toml' docs

michaeljones:
Written for gleam 0.12 which is the current release though changes will be needed for gleam 0.13.

Putting this up for discussion as I'm not confident in where this should be placed or what style of writing to use. Happy to change any and all of it.


lpil (approved):
Fab, thank you.

PR #60, opened 2020-11-21
Style `code` in lists too

michaeljones:
Currently the code styling is limited to 'p' tags but 'li' tags don't have a nested 'p' tag so we need to account for them too.


lpil (approved):
Nice, thanks. How did you spot that?

PR #61, opened 2020-11-22
Move types & optionality to new lines

michaeljones:
Otherwise the headers become quite long and wrap awkwardly on mobile screens.

Currently:
image

With change:
image

Sorry the images come out so large.


lpil (approved):
Much better, thanks.

PR #62, opened 2020-11-22
Make 'Note' bold

michaeljones:
Minor and subjective change. Our 'bold' doesn't pop too much so this just serves to separate the 'Note:' label from its contents a little for clarity.

Sorry this is so minor. It is just a personal preference of mine that I should have included in the first pass. Feel free to ignore if you have a preference against it.


lpil (approved):
Ta!

PR #63, opened 2020-11-22
[v0.13] Include 'repository' field in gleam.toml docs

michaeljones:
Designed to be merged when 0.13 is release so it is on a `0.13-changes` branch. Happy to rename if you'd like to take another approach. Just wanted this to be set and ready.

I'll rebase so it only includes the second commit if you merge the first in the other PR. (And if you don't think I'll remove it from this one.) I just figure they'll probably clash so leaving it here seems reasonable.

### DIFF
--- a/writing-gleam/gleam-toml.md
+++ b/writing-gleam/gleam-toml.md
@@ -31,6 +31,22 @@ A description of your project.
 **Note**: This does not determine the description on the hex.pm project page when publishing your
 project. That is determined by the `description` entry in the `src/*.app.src` file in your project.
 
+## `repository`
+
+`object` - *optional*
+
+Specifies the online source repository for this project's code.
+
+This enables source links, in the generated documentation, from types, constants & functions to
+their defining lines of code in the repository.
+
+```toml
+repository = { type = "GitHub", user = "example", repo = "project" }
+repository = { type = "GitLab", user = "example", repo = "project" }
+repository = { type = "BitBucket", user = "example", repo = "project" }
+repository = { type = "Custom", url = "https://repo.example.com" }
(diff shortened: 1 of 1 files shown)


lpil (commented):
Fab, thank you.

michaeljones:
This does make me wonder if we should remove the `Custom` entry as we don't do anything with it really as we can't predict the source link format. That, or we can auto-generated an entry in the `links` in the side bar for all of these including `Custom` which links back to the repository. Saves the user adding it to `links` themselves though I'm not sure what to do if they don't want it.

I would also be slightly happier if the config file expected lower case types: `github`, `bitbucket`, etc, rather than `GitHub` and `BitBucket`. Purely subjective aesthetics though.

What do you think?

lpil:
I think we should keep it so we can still render a link to the repository in the sidebar. Even if we can't give specific line numbers it's still useful for people to be able to get to the code.

michaeljones:
Cool, but we don't have the code to generate a link automatically from the repository yet. Shall I make a ticket for that?

lpil:
Yes please, that would be great￼

michallepicki:
Just a note - maybe it could be done in the same PR (?) - I think the `version` field will have to be marked as required in the same file for v0.13 as well.

michaeljones:
Smart! I should have realised that.

lpil:
Hold off for now, I think I'm going to try and find a way to do this is a backwards compatible way for the next release. Taking the version from the .app rather than comparing them perhaps.

I want it to be as friendly as possible!

lpil (approved):
Thank you

PR #64, opened 2020-11-25
Provide links to toml.io

michaeljones:
Pretty subjective. I can understand if you consider it overkill. Especially with the repetition but I kind of think that within reason the more explanation & context the better.

---

To make it easier for beginner's to understand the reference.

Based in the fact that tables in toml can be flattened out to multiple
lines of variables and given that that syntax is used in some gleam
projects, it feels like it makes sense to draw attention to the
potential representation.

### DIFF
--- a/writing-gleam/gleam-toml.md
+++ b/writing-gleam/gleam-toml.md
@@ -3,8 +3,10 @@ title: gleam.toml
 layout: page
 ---
 
-All Gleam projects require a `gleam.toml` configuration file. The configuration file allows you to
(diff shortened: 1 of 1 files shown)


lpil (commented):
I like the link to the toml reference, very useful. I'm not so sure about the notes on what a table is, it seems like that would get repeated often and not be useful to most people.￼

michaeljones:
Updated.

lpil (approved):
Thanks @michaeljones !

PR #66, opened 2020-11-30
Add Language Tour to end of Getting Started.

rockerBOO:
I am just starting gleam and wanted to add this. I read the getting started and then I had to scroll up to find the language tour part. But it wasn't on this page anywhere and had to go to the previous page. 

This adds it at the very bottom so those going through the same flow can get a next step.


lpil (approved):
Hey, much nicer! Thank you

PR #74, opened 2021-01-06
Provide an escript main function signature in the documentation

mlapping:
Update the documentation with an escript main function signature so that it's easier to work with arguments as strings rather than charlists. This took me entirely too long to figure out, hopefully it helps someone else.


lpil (approved):
Thank you, much clearer

PR #76, opened 2021-01-08
Added gleam highlighting to highlight.js

itsgreggreg:
Not sure you're interested in this but I added proper syntax highlighting to the gleam snippets in the book.

I'm sure there's a better way to load this module but I'm not super familiar with including JS in mdbook. So `highlight.js` has been overridden directly to include the gleam highlighter.

A proper module for highlightjs is available in npm https://www.npmjs.com/package/highlightjs-gleam in case anyone knows how to properly load it in mdbook.

Here's some examples:

Screen Shot 2021-01-07 at 7 55 04 PM
Screen Shot 2021-01-07 at 7 51 23 PM
Screen Shot 2021-01-07 at 7 52 42 PM

### DIFF
--- a/book/searcher.js
+++ b/book/searcher.js
@@ -145,6 +145,11 @@ window.search = window.search || {};
             url.push("");
         }
 
+        // encodeURIComponent escapes all chars that could allow an XSS except
+        // for '. Due to that we also manually replace ' with its url-encoded
+        // representation (%27).
+        var searchterms = encodeURIComponent(searchterms.join(" ")).replace(/\'/g, "%27");
+
         return '<a href="' + path_to_root + url[0] + '?' + URL_MARK_PARAM + '=' + searchterms + '#' + url[1]
             + '" aria-details="teaser_' + teaser_count + '">' + result.doc.breadcrumbs + '</a>'
             + '<span class="teaser" id="teaser_' + teaser_count + '" aria-label="Search Result Teaser">' 
--- a/book-src/tour/assert.md
+++ b/book-src/tour/assert.md
@@ -19,7 +19,7 @@ prototype application, so we want to only spend time on the success path for now
 For these situations Gleam provides `assert`, a keyword that causes the
 program to crash if a pattern does not match.
 
-```rust,noplaypen
+```gleam
 assert Ok(i) = parse_int("123")
 i // => 123
 ```
@@ -28,7 +28,7 @@ Here the `assert` keyword has been used to say "this function must return an
 `Ok` value" and we haven't had to write any error handling. The inner value
 is assigned the variable `i` and the program continues.
 
-```rust,noplaypen
+```gleam
 assert Ok(i) = parse_int("not an int")
 ```
 
--- a/book-src/tour/bit-strings.md
+++ b/book-src/tour/bit-strings.md
@@ -3,7 +3,7 @@
 Gleam offers a syntax for working directly with raw data in the form of bit
 strings.
 
-```rust,noplaypen
+```gleam
 // A bit string of the 8 bit int value 3
 <<3>>
 
--- a/book-src/tour/bools.md
+++ b/book-src/tour/bools.md
@@ -4,7 +4,7 @@ A Bool can be either `True` or `False`.
 
 Gleam defines a handful of operators that work with Bools.
 
-```rust,noplaypen
+```gleam
 False && False // => False
 False && True  // => False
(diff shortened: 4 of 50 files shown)


book/highlight.js
```diff
@@ -3,4 +3,5586 @@
...
...2ps vfmsubadd132pd vfmsubadd312ps vfmsubadd312pd vfmsubadd213ps vfmsubadd213pd vfmsubadd123ps vfmsubadd123pd vfmsubadd231ps vfmsubadd231pd vfmsubadd321ps vfmsubadd321pd vfnmadd132ps vfnmadd132pd vfnmadd312ps vfnmadd312pd vfnmadd213ps vfnmadd213pd vfnmadd123ps vfnmadd123pd vfnmadd231ps vfnmadd231pd vfnmadd321ps vfnmadd321pd vfnmsub132ps vfnmsub132pd vfnmsub312ps vfnmsub312pd vfnmsub213ps vfnmsub213pd vfnmsub123ps vfnmsub123pd vfnmsub231ps vfnmsub231pd vfnmsub321ps vfnmsub321pd vfmadd132ss vfmadd132sd vfmadd312ss vfmadd312sd vfmadd213ss vfmadd213sd vfmadd123ss vfmadd123sd vfmadd231ss vfmadd231sd vfmadd321ss vfmadd321sd vfmsub132ss vfmsub132sd vfmsub312ss vfmsub312sd vfmsub213ss vfmsub213sd vfmsub123ss vfmsub123sd vfmsub231ss vfmsub231sd vfmsub321ss vfmsub321sd vfnmadd132ss vfnmadd132sd vfnmadd312ss vfnmadd312sd vfnmadd213ss vfnmadd213sd vfnmadd123ss vfnmadd123sd vfnmadd231ss vfnmadd231sd vfnmadd321ss vfnmadd321sd vfnmsub132ss vfnmsub132sd vfnmsub312ss vfnmsub312sd vfnmsub213ss vfnmsub213sd vfnmsub123ss vfnmsub123sd vfnmsub231ss vfnmsub231sd vfnmsub321ss vfnmsub321sd rdfsbase rdgsbase rdrand wrfsbase wrgsbase vcvtph2ps vcvtps2ph adcx adox rdseed clac stac xstore xcryptecb xcryptcbc xc
```
itsgreggreg, line 5484: all of the code above this line is the formatted source of what used to be `highlight.js` it's "minified" just formatted. This is so the source below can be formatted without having to jump through any hoops. The code below this line is the highlighter for Gleam source. Perhaps we can minify this whole file if we don't need to be able to modify the gleam highlighter. But more importantly there's probably a better way to load all of this with mdbook.

lpil: It looks like we can add additional JavaScript in the mdbook config https://github.com/budziq/mdBook/blob/master/tests/tomlconfig.rs#L192

Perhaps the way to go is to reset this file to the minified state and then to add the Gleam specific code as a new file. I worry about this file being overwritten by the book as it is currently.

theme/highlight.js
```diff
@@ -0,0 +1,5588 @@
+/*
```
lpil, line 1: Is this the same file included again?

lpil:
This is fantastic, thank you! I've been wanting highlighting for some time.

book/highlight.js
itsgreggreg: righteous, got it working without the override 👍

theme/highlight.js
itsgreggreg: yeah, i put the comment in the wrong file. there will be 2 copies of the highlighter file, one in `/javascript` which is the source and one in `/book/javascript` which is generated by `mdbook build`. Happy to move the file somewhere else if we don't want it in `/javascript`

book/searcher.js
```diff
@@ -145,6 +145,11 @@ window.search = window.search || {};
             url.push("");
         }
 
+        // encodeURIComponent escapes all chars that could allow an XSS except
+        // for '. Due to that we also manually replace ' with its url-encoded
+        // representation (%27).
+        var searchterms = encodeURIComponent(searchterms.join(" ")).replace(/\'/g, "%27");
+
```
itsgreggreg, line 152: no idea what's going on here, maybe i have a newer version of mdbook? happy to remove if we don't want

lpil: Looks like it. Let's keep it

lpil (approved):
Thanks!

PR #79, opened 2021-01-14
Adds a RSS feed for news

### LINKED ISSUE #78: RSS feed for news
👋 Hello!

I've been interested in Gleam for a while and would like to keep myself in the loop around new releases. My normal way to do that would be by subsribing to the RSS feed for news.

However, there is no RSS feed there 😢  Would you be happy to have one? I think there is a plugin for GitHub pages that generates one... I'd be happy to PR if this seems like a good idea.

gampleman:
Closes #78.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -5,11 +5,17 @@ markdown: kramdown
 
 permalink: "/news/:title/"
 
+title: Gleam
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you

PR #80, opened 2021-01-25
Document string escapes

itsgreggreg:
Screenshot_2021-01-25 Strings - The Gleam Book


itsgreggreg:
oops bad merge one sec

PR #82, opened 2021-01-25
Fix typo in Elixir operators

jraregris:
`<=` was written as `=<` in the table comparing Elixir and Gleam operators.

In Elixir `=<` gives us SyntaxError, while `<=` is the "Less or equal" operator.


lpil (approved):
Thanks! Must have been copied from the Erlang one

PR #83, opened 2021-01-26
Expanded Bit Strings page

itsgreggreg:
Screenshot_2021-01-25 Bit strings - The Gleam Book


lpil (approved):
Oh excellent! Thank you very might. This might actually be the clearest docs I've read on this subject, lol

PR #85, opened 2021-01-29
BitString docs to keep up with the stricter rules

itsgreggreg:
Keeping up with the changes in this PR https://github.com/gleam-lang/gleam/pull/953
Screenshot_2021-01-29 Bit strings - The Gleam Book


lpil (approved):
Wonderful, thank you! Much better than the other docs on bitstrings I've been trying to wrap my head around

PR #86, opened 2021-01-30
Fix typo in Elixir cheatsheet

malcolmseyd:
In the Atoms section, "boolean" was written as "boolen".


lpil (approved):
Good catch! Thank you

PR #88, opened 2021-02-20
timcondit patch 1

timcondit:
Interesting language, and good docs!

(And I know this isn't the place for it, but why is the `tuple` type lowercase? Seems like other types are all capitalized.)

### DIFF
--- a/book-src/tour/custom-types.md
+++ b/book-src/tour/custom-types.md
@@ -39,10 +39,10 @@ fn cats() {
 Custom types in Gleam can be defined with multiple constructors, making them a
 way of modeling data that can be one of a few different variants.
 
-We've already seen a custom type with multiple constructors in the Language Tour,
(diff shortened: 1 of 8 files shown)


book-src/tour/custom-types.md
```diff
@@ -52,13 +52,13 @@ pub type Bool {
 }
 ```
 
-It's a simple custom type which constructors that take no arguments at all!
+It's a simple custom type with a constructor that takes no arguments at all!
```
lpil, line 55:
```suggestion
It's a simple custom type with constructors that takes no arguments at all!
```

lpil (approved):
Wonderful, thank you.

The `tuple` type constructor is lowercase as it is the only type constructor that accepts a variable number of arguments.

This syntax may change in the near future though: https://github.com/gleam-lang/suggestions/issues/92

PR #90, opened 2021-02-21
Document binary, octal and hexadecimal ints

varnerac:


lpil (approved):
Fantastic, thank you!

PR #91, opened 2021-02-22
Reference string.append, string.concat, StringBuilder

htstudios:
Ref: https://github.com/gleam-lang/gleam/issues/985

### DIFF
--- a/book-src/tour/strings.md
+++ b/book-src/tour/strings.md
@@ -48,3 +48,16 @@ Similarly all backslash characters must be escaped:
 // A Decorative border /\/\/\/\
 "/\\/\\/\\/\\"
 ```
+
+### String Operations
+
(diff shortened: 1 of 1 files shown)


lpil:
I'm not sure this is the right place to put it because it is using language concepts that have not yet been introduced at this point in the tour. It would need to come after functions and imports. Perhaps a "Gleam Cookbook" would be the way to go here.

htstudios:
If you changed your mind on this. Let me know.

I would put a chapter about the stdlib on the book and only reference some very common things... string operations like this will be looked up by a lot of people onboarding, especially if gleam grows and the devs are of all flavours and skill levels ;)

lpil:
You are certainly right that we are missing lots of stuff here. We'll tackle this problem in the near future!

PR #92, opened 2021-02-23
Fix repository.type examples in gleam.toml

rjdellecese:
See https://github.com/gleam-lang/gleam/blob/a474ef64f67b16cf98e8e7b1e2d71819152a7a3a/src/config.rs#L42-L50.


lpil (approved):
Thank you

PR #93, opened 2021-02-23
Correct docs commands examples

rjdellecese:
- `--version` is a required arg for the `gleam docs build` command
- The value for `--version` must match the value in `.app.src`, which means that it must be valid semver, which means (I think) that it must not have a `v` preceding it... or at least I can't find any examples of that version format in the [`rebar3 docs`](https://rebar3.org/docs/package_management/publishing-packages/) or the [`hex.pm docs`](https://hex.pm/docs/rebar3_publish), and so it would be atypical.


lpil (approved):
Thank you!

PR #95, opened 2021-03-26
Rust Cheatsheet: Use the assigned value

michaeljones:
I'm not sure if this is valid rust but the way the doc reads, it feels like this is the intent.


lpil (approved):
Oh yes, much better. Thank you

PR #96, opened 2021-03-30
Add discussions on Github.

CrowdHailer:
Rather controversially I've removed the Reddit link. There was discussion about not having too many separate places to check, and it would potentially encourage more people to try the discussions if no reddit link, so we'll find out sooner if discussions are a good idea.


lpil (approved):
Makes sense, thank you

PR #97, opened 2021-04-03
Update target tag

michaeljones:
I guess we should find a way to do this automatically but I'm not
familiar with any scripts in the website set up that might be well
placed to do that.


lpil (approved):
Thank you. We can add a note to the release instructions if nothing else

https://github.com/gleam-lang/gleam/blob/main/RELEASE.md

PR #105, opened 2021-05-05
Add note about gleam/bool in the book

tranzystorek-io:
I remember being confused about bool operators that are typical in other languages. This should clarify things up a bit more for any newcomers 😄


lpil (approved):
Thank you

PR #107, opened 2021-05-12
Fix anonymous function call in Elixir guide

sneako:
These gleam-for-x guides are really a great thing to have!

I noticed this small issue with one of the Elixir examples


lpil (approved):
Ah! Thank you

PR #110, opened 2021-06-04
Add link to Elm cheatsheet & sort list

michaeljones:
Not sure if you want the list sorted. I can remove that commit if preferable. Just seems like it is getting to a length where an order is sensible.


lpil (approved):
Love it, thanks

PR #113, opened 2021-06-05
Further adjustments to elm cheatsheet

michaeljones:
Attempting to tighten up the wording in the overview section which I
felt was a bit hard to parse in places.

Add a few more bits of information and a few more links out to external
references.

Put more of the types in backticks as a general formatting style.

### DIFF
--- a/cheatsheets/gleam-for-elm-users.md
+++ b/cheatsheets/gleam-for-elm-users.md
@@ -726,7 +726,7 @@ pub type Result(value, reason) {
 
 The standard library provides the [gleam/result](https://hexdocs.pm/gleam_stdlib/gleam/result/) module for interacting with `Result` values.
 
-Gleam has a `try` keyword that allows for early exit from a function if a `Result` is an error. The equivalent in Elm would require the use of `Result.andThen`. The `try` keyword in Gleam provides syntactic sugar which simplifies functions that handle results.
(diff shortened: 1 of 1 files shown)


cheatsheets/gleam-for-elm-users.md
```diff
@@ -719,9 +724,9 @@ pub type Result(value, reason) {
 }
 ```
 
-The standard library provides the [gleam/result](https://hexdocs.pm/gleam_stdlib/gleam/result/) module for interacting with result values.
+The standard library provides the [gleam/result](https://hexdocs.pm/gleam_stdlib/gleam/result/) module for interacting with `Result` values.
 
-Gleam has a `try` keyword that allows for early exit from a function if a Result is an error. The equivalent in Elm would require the use of `Result.andThen`. The `try` keyword in Gleam provides syntactic sugar which simplifies functions that handle results.
+Gleam has a `try` keyword that allows for early exit from a function if a `Result` is an error. The equivalent in Elm would require the use of `Result.andThen`. The `try` keyword in Gleam provides syntactic sugar which simplifies functions that handle results.
```
lpil, line 729:
```suggestion
Gleam has a `try` keyword that allows for early exit from a block if a `Result` is an error. The equivalent in Elm would require the use of `Result.andThen`. The `try` keyword in Gleam provides syntactic sugar which simplifies functions that handle results.
```

lpil (approved):
Looks grand, thank you

PR #114, opened 2021-06-05
Fix errors/typos in gleam-for-elm

jfmengels:
Hi there :wave: 
Thanks a lot for this great guide @michaeljones. I noticed a few errors/typos so I thought I'd make a PR.

Additionally, I noticed a few problems that I didn't fix, which I've described below:

> It is recommended to use records for more larger numbers of entries.

This sentence reads weird, but I didn't suggest an alternative because I figured that could be rejected and could delay merging this PR.

> You can pattern match on lists in Elm, but only in case-statements.

While I think this is useful to know, the example afterwards demonstrates list construction, not pattern matching, which is confusing. Maybe talk about the `::` operator right before the example?

> Elm uses type aliases to define the layout of records. Gleam uses Custom Types to achieve a similar result.

Elm uses type aliases for other things than records. I understand from this that Gleam type aliases can only be for records? Is that right, or could this be phrased differently?

> Javascript

It's JavaScript with a capital S. I didn't want to change this everywhere for this PR.

> It is not possible to publish an Elm package that includes Javascript script

script -> code?

### DIFF
--- a/cheatsheets/gleam-for-elm-users.md
+++ b/cheatsheets/gleam-for-elm-users.md
@@ -242,9 +242,9 @@ Elm has no built in way to label arguments. Instead it would standard for a func
 
 ```elm
 defaultOptions =
-  { inside : defaultString
-  , each : defaultPattern,
-  , with : defaultReplacement
+  { inside = defaultString
+  , each = defaultPattern,
+  , with = defaultReplacement
   }
 
 replace opts =
@@ -252,9 +252,9 @@ replace opts =
 ```
 
 ```elm
-replace { each: ",", with: " ", inside: "A,B,C" }
+replace { each = ",", with = " ", inside = "A,B,C" }
 
-replace { defaultOptions | inside: "A,B,C,D" }
+replace { defaultOptions | inside = "A,B,C,D" }
 ```
 
 #### Gleam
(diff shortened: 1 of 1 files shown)


michaeljones (approved):

michaeljones:
Thank you for proofreading! Slightly embarrassing but I'm used to not being the best at spotting my own mistakes. Kind of you to take the time to improve it.

Yeah, I imagine that:

> It is recommended to use records for ~~more~~ larger numbers of entries.

Might read better and be ok? 

Your point about "You can pattern match on lists in Elm, but only in case-statements" is totally fair. It is possible that the section should be changed in focus a bit. It is a hang over from the Elixir cheatsheet that I started with as the base and it makes sense as a section there but not so much with the examples here. I guess the guide should have pattern matching examples but these aren't great.

Good point about the type aliases. I didn't think that through. Makes sense to change that some how.

I had never internalised the need for the capital S in JavaScript. Interesting! I guess that can be done. I can understand not doing it in this PR though. 

> script -> code?

Yup! 

Thanks again.

michaeljones:
The changes here don't seem to clash with the ones in the other branch/PR so we can merge any either order.

lpil (approved):
Thank you!￼

PR #115, opened 2021-06-05
Elm cheatsheet tweaks

michaeljones:
Based on the feedback from @jfmengels. 

Hopefully all reasonable.


lpil (approved):
Good stuff, thank you

PR #118, opened 2021-06-10
Add Highlight JS based syntax highlighting of Gleam code

JohnDoneth:
Resolves https://github.com/gleam-lang/gleam/issues/981

Updates all Gleam specific code blocks to use the Gleam specifier while taking care to avoid changing the Rust specific code blocks to Gleam. Uses the same light theme CDN and highlight JS rules as the docs.

This also affects the three code snippets on the main page.

----

### Before
Screenshot 2021-06-09 at 22-28-48 Gleam v0 15 released – Gleam

### After
image


lpil (approved):
Thank you

PR #121, opened 2021-07-18
Fix language tour code highlighting

shunjilin:
Language tour is using v 10.1.1 of highlightjs which does not support highlightAll function, use initHighlightingOnLoad instead as a quick fix

may want to consider upgrading to the same version and using the cdn solution like in the rest of the docs, if I am not wrong that would mean putting the script tag in every html page?


lpil (approved):
Sounds good to me!

PR #122, opened 2021-08-03
Correct typos

ShalokShalom:
Other data structes like Lists, and also Tuples itself at other places, are always refered to in plural.


lpil (approved):
Thank you

PR #123, opened 2021-08-03
Clearify wildcard

ShalokShalom:
So, if I assume correctly - this comes not out clearly of the text - then is the last entry in a case expression always the wildcard, yes?

So in case I put xyz there, this is the 'matches all' case, correct?

In case that is so, let us mention this specifically, since its unique among all languages, so far as I know at least.

I also find it confusing, that different values, like n, _other, other and _ are used in the examples.

I guess this should suggest, that I can use everything, while it also suggests a certain 'conventional pattern' 

But in that case, its more than one convention, and actually mixing multiple different ones.

### DIFF
--- a/book-src/tour/case-expressions.md
+++ b/book-src/tour/case-expressions.md
@@ -6,14 +6,14 @@ _pattern matching_.
 
 Here we match on an `Int` and return a specific string for the values 0, 1,
 and 2. The final pattern matches any other value that did not match any of
-the previous patterns.
+the previous patterns. You can name this entry, as you want. 
 
 ```gleam
 case some_number {
   0 -> "Zero"
   1 -> "One"
   2 -> "Two"
-  _ -> "Some other number" // This matches anything
+  n -> "Some other number" // This matches anything
 }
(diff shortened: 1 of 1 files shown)


lpil:
Giving a discarded variable a descriptive name is often useful for clarify. I would prefer to keep them in so that users know they can give them names.

book-src/tour/case-expressions.md
```diff
@@ -5,15 +5,15 @@ allows us to say "if the data has this shape then do that", which we call
...
+and 2. The final pattern matches any other value that did not match any of
 the previous patterns.
 
 ```gleam
 case some_number {
   0 -> "Zero"
   1 -> "One"
   2 -> "Two"
-  n -> "Some other number" // This matches anything
+  _ -> "Some other number" // This matches anything
```
lpil, line 16: Perhaps we could use the variable `n` in the expression here?

lpil:
I think the existing code is in good in that it shows the user how they can name discarded variables with examples.

PR #124, opened 2021-08-09
Add info on unions and matching to Rust cheatsheet

tranzystorek-io:
Adds examples that illustrate `case` usage since it's almost identical to `match`


lpil (approved):
Thank you

PR #128, opened 2021-10-22
Update result.html

inoas:
Looks different... copy&paste accident?

https://gleam.run/book/tour/custom-types.html#custom-types
vs
https://gleam.run/book/tour/result.html#resultvalue-error


inoas:
This is bollox.

PR #135, opened 2021-12-08
Fix issues with the command line docs

michaeljones:
I believe these are legitimate changes but I might be wrong.


lpil (approved):
Oops! Thank you. Some sloppy editing from me while tired there.

PR #137, opened 2021-12-25
Use version v0.18.2 instead of v0.14.4 in the build-from-source docs

Dobiasd:
This could help some avoid confusion some beginners, following the guide and then wondering why `gleam` does not even have the expected subcommands. :slightly_smiling_face:


lpil (approved):
Thank you

PR #139, opened 2022-02-09
Fix syntax error in the snippet

slashmili:
I was trying to copy&past the snippet but got into a trouble and the compiler complained about the parentheses


lpil (approved):
Thank you <3

PR #144, opened 2022-02-26
Wait for processes to exit in multi-thread example

mcastorina:
The previous version wasn't waiting for the processes to exit, so nothing was getting printed.

I'm still very new to gleam (which is why I was running the examples on the website), but hopefully this change is helpful!


lpil (approved):
Thank you!

PR #155, opened 2022-03-14
Explain the spread syntax used in pattern matching.

hayleigh-dot-dev:
I’ll build this and fix any inevitable typos when I’m at a computer and it’s not 2am!


lpil:
Ready to review?

PR #159, opened 2022-04-12
fix: Sponsors button background

HarryET:
Sponsors button has a black background, as seen below:
image
This fix removes the use of an `iframe`, taking that logic/element into the main layout. It also imports the GitHub CSS. The button now looks like this:
image


lpil (approved):
Thanks!

PR #160, opened 2022-04-13
fix: Page stuck because of Github's CSS

HarryET:
Put CSS into the project and remove the body modifier, no longer requires arbitory css from GitHub


lpil (commented):
These classes are very generic and we don't seem to be using the majority of them. I think we need to trim these down. I'll revert to the the officially supported iframe embed for now, we can add a trimmed down version of the same widget later

HarryET:
> These classes are very generic and we don't seem to be using the majority of them. I think we need to trim these down. I'll revert to the the officially supported iframe embed for now, we can add a trimmed down version of the same widget later

Yeah probably for the best, I'll close this too and maybe its worth making something more custom at some point. Sorry for the issues 😓

PR #162, opened 2022-04-14
fix sponsor button

herminiotorres:


javascript/sponsor.js
```diff
@@ -0,0 +1,15 @@
...
+
+function scaleIncreaseHeart() {
+  let svg = sponsorLink.getElementsByTagName('svg')[0];
+  svg.setAttribute('transform', `scale(1)`);
+}
+
+function scaleDecreaseHeart(){
+  let svg = sponsorLink.getElementsByTagName('svg')[0];
+  svg.setAttribute('transform', `scale(1.2)`);
+}
```
lpil, line 15: Is this done because we can't use `:hover` CSS?

herminiotorres: I use this idea because when you have this element, like `a/button`, and you need to scale only the heart when the entire element area has hovered, only adding events is possible otherwise it will scale the entire element.

lpil: Sorry, perhaps I'm getting confused. Wouldn't a selector like `a:hover svg` work here?

lpil (approved):
Thank you!

PR #166, opened 2022-04-30
Update section about negation of `Bool`s

majjoha:
Thank you for a great project. I've eagerly been following the development of Gleam for a while. :-)

While reading through the language tour, I noticed that the section about booleans is outdated, as we now have the `!` operator available, and I thought it was worth updating it with a code example.


lpil (approved):
Thank you

PR #167, opened 2022-04-30
Change gleam to Gleam

majjoha:
The section about expression blocks contains a minor typo, as gleam should be written Gleam which this pull request fixes.


lpil (approved):
Thank you

PR #168, opened 2022-05-01
pipe and capture chapter

inoas:


inoas:
Attemps https://github.com/gleam-lang/website/issues/25

inoas:
This may be obvious but for newbies/starters this may help:

TODO: pipe into anon function
TODO: pipe into case

inoas:
TODO: Shorten language.

inoas:
Any idea if wie should and/how we can add examples of railway oriented programming <https://vimeo.com/113707214>, either-monad?

TODO: Add remark about <https://redrapids.medium.com/learning-elixir-its-all-reduce-204d05f52ee7>

inoas:
TODO show auto capturing of left over args after setting labels:
https://johndoneth.github.io/gleam-playground/?s=A4VwRgBAZgdhDO54BcBOBDAxsiAKAtgJYwgCmMAJhOgFwQCSMyANAuGugBblVh2PIAlBAC0APgZMIAbwBQEBdVEQw8iAF9Zs0JFgR86YrmFyFARggAfCYjAoM2XLY7dKdAEzCA9F9Fm1FtZsdhyORCQ8Ht6%2BIv6aQA%3D%3D

lpil (commented):
The writing style here is quite different to the rest of the tour. I think we should keep the concise style as used before. This could be good in a more detailed book though.

What new information has been added here other than the text changes? I can't quite tell because a lot of things have been changed in one PR. Does it cover a feature that wasn't before?

inoas:
> What new information has been added here other than the text changes? I can't quite tell because a lot of things have been changed in one PR. Does it cover a feature that wasn't before?

These are the additions/changes:

- Moved the 2nd level chapters about piping and function capturing from `functions` into this new chapter.
- Added comparision to method chaining, well known and prevlanet in OO languages to make users feel at home.
- Added reference to unix style piping as that kind of returns `Result`s
- Replaced `iodata.*` with `string_builder` reference
- Made more obvious the steps between parenthesis closure, intermediate variables and piping
- Explained why during piping function calls with arity of 1 may be written without parenthesis which may - if left unexplained - confuse new users
- Added examples on pipes with multiple arguments where the other args have to be supplied during piping
- Explained that during piping types result and argument types match in Gleam (other than in say Elixir where only patterns have to match AFAIU)
- Explained that functional piping does not mutate the data and thus is relativly save to spawn async processes within itself unlike in OO method chaining
- Explained Gleams feature to call the returned function passing in the argument if the function given in a pipe step is already complete
- Explained that the capture symbol can be used to push return values into specific argument positions during piping
- Explained that Gleam will autodetect the argument whole during piping if labels are give for all the other args
- Explained that you can also pipe into anonymous functions

inoas:
Let me know if I can do anything here, it isn't urgend, but I can spend some time next month if there is any chance you want this or parts of this. I would also love to collaborate on improvements and reviews if someone wants to join.

lpil:
I'm sorry @inoas but I think this is too big a change and not in keeping with the existing style of the rest of the tour, especially with all the information about other programming paradimes.

I think the best approach would be to revert to the previous text and then add any additions for functionality that it did not explain.

I also don't want to have a joint pipe and function capture syntax page as it implies they are related in some fashion, when instead they are two independant features.

inoas:
@lpil that's fair. I will keep this open as a draft and once I have some time I can will make small PRs.

inoas:
I don't think this is required anymore.

PR #169, opened 2022-05-02
Add tuple type explanation

inoas:


lpil (commented):
I don't think we should be explaining the rationale for each design design in the language in the tour, it is only a high level overview.

This is more suited to a book, conference talk, or other place for trivia.

inoas:
My stance is that the hardest part about learning new languages is learning quirks/irregularities. This is certainly unexpected from a newbie point of view.

I can cut down on the details and try to find a wording that just explains what the type contructor is for tuples instead of explaining why.

I think that would be helpful. Would that work for you?

lpil:
This is definitely not the place for this kind of information. Everything in the language tour is something that you need to know to write Gleam programs. I suspect the majority of Gleam programmers in the future will never learn this and won't be impacted by it.

This is not something that is documented in the intro docs any other language I have come across, and having a different syntax for certain types such as functions, tuples, lists, and maps is very common.

inoas:
Ok, but I tried to add a test with a custom type and added `Tuple()` in there and it did not compile. How should I know better as a starter?

I understand that this is only a language tour, but we lack a complete written documentation (will it be free?) so this was the best place I thought this could go into.

Let me know if you want me to do anything about it the way you prefer and I will reopen.

PR #172, opened 2022-05-21
Small changes to comparison with Elixir

josevalim:
* The namespace is between modules and variables
* Clarity interoperability in regards to distribution

The FAQ is typically really clear but it was vague which parts the interoperability history is different. Here is a proposal to go with more concrete examples.

### DIFF
--- a/frequently-asked-questions.md
+++ b/frequently-asked-questions.md
@@ -152,11 +152,10 @@ Here's a non-exhaustive list of differences:
 - Elixir has a powerful macro system, Gleam has no metaprogramming features.
 - Elixir's compiler is written in Erlang and Elixir, Gleam's is written in Rust.
 - Elixir uses Ruby style syntax, Gleam has a C family style syntax.
-- Elixir code can be difficult to use from other BEAM languages, Gleam code is
-  designed to be usable from all BEAM languages.
-- Elixir has a namespace for module functions and another for anonymous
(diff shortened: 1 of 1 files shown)


frequently-asked-questions.md
```diff
@@ -152,11 +152,10 @@ Here's a non-exhaustive list of differences:
...
 - Elixir uses Ruby style syntax, Gleam has a C family style syntax.
-- Elixir code can be difficult to use from other BEAM languages, Gleam code is
-  designed to be usable from all BEAM languages.
-- Elixir has a namespace for module functions and another for anonymous
-  functions, Gleam has one unified namespace (so there's no special `fun.()`
-  syntax).
+- Elixir has a namespace for module functions and another for variables,
+  Gleam has one unified namespace (so there's no special `fun.()` syntax).
+- Gleam standard library is distributed as Hex packages, which makes interoperability
+  with other BEAM languages easier.
```
josevalim, line 158: We could also mention how Elixir does things:

```suggestion
- Elixir standard library is part of its distribution and typically managed at the Operating
  System level. Gleam standard library is distributed as Hex packages, which makes
  interoperability with other BEAM languages easier.
```

But I think focusing on how Gleam is better is the cleanest.

lpil (approved):
Much better, thank you

PR #176, opened 2022-06-21
Update gleam-for-elixir-users.md

cmcaine:
Cleared up language around variable assignments and made the pattern matching slightly more realistic so that the examples there (arguably) make more sense.

### DIFF
--- a/cheatsheets/gleam-for-elixir-users.md
+++ b/cheatsheets/gleam-for-elixir-users.md
@@ -81,7 +81,7 @@ size = 1
 
 #### Gleam
 
-Gleam has the `let` keyword before its variable names.
+Gleam has the `let` keyword before each variable assignment.
 
 ```gleam
 let size = 50
@@ -94,8 +94,9 @@ let size = 1
 #### Elixir
 
 ```elixir
-[x] = [1] # assert `x` is a 1 element list and assign it to this element's value
-2 = x # runtime error because x's value is 1
+[x, y] = [1, 2]
+2 = y # assert that y is 2
+2 = x # runtime error
 [y] = "Hello" # runtime error
 ```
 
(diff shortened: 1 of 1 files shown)


cheatsheets/gleam-for-elixir-users.md
```diff
@@ -94,8 +94,9 @@ let size = 1
 #### Elixir
 
 ```elixir
-[x] = [1] # assert `x` is a 1 element list and assign it to this element's value
-2 = x # runtime error because x's value is 1
+[x, y] = [1, 2]
+2 = y # assert that y is 2
+2 = x # runtime error
```
cmcaine, line 99:
```suggestion
[x, y] = [1, 2] # assign 1 to x and 2 to y by pattern matching on the list [1, 2].
2 = y # assert that y is 2
2 = x # runtime error because x's value is 1
```

Here's a suggestion for adding comments as you asked. I'm fairly sure that the old comment:

```
[x] = [1] # assert `x` is a 1 element list and assign it to this element's value
```

is just wrong (`[x] = [1]` doesn't make any assertions about the current value of x, and doesn't cause x to be a list), so I don't think it should be restored.

lpil: Ah yes, it should "assert the value is a 1 element list"

cmcaine: How about this?

```suggestion
[x, y] = [1, 2] # assert that the value is a list of length 2 and assign the first element to x and the second to y
2 = y # assert that y is 2
2 = x # runtime error because x's value is 1
```

cheatsheets/gleam-for-elixir-users.md
```diff
@@ -104,7 +105,8 @@ let size = 1
 In Gleam, `let` and `=` can be used for pattern matching, but you'll get compile errors if there's a type mismatch, and a runtime error if there's a value mismatch. For assertions, the equivalent `assert` keyword is preferred.
 
 ```gleam
-let [x] = [1]
+let [x, y] = [1, 2]
```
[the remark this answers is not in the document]
lpil: Yup. Using let here will not be permitted once we have full exhaustiveness checking

cheatsheets/gleam-for-elixir-users.md
lpil: Sounds good!

cmcaine:
Lgtm

lpil (approved):
Brilliant! Thanks for working through the finer details with me here

PR #177, opened 2022-06-27
Advertise TCO feature :)

inoas:
Ref https://github.com/gleam-lang/stdlib/issues/321#issuecomment-1167794314


lpil:
I don't think this is the right place for this as it is not a major feature that would draw people to the language. I think this would be better in best practices documentation that we could have in future.

inoas:
Where would we collect topics/ideas for that best practises documentation?

PR #181, opened 2022-10-13
Rename modulo to remainder, add notice for Python users how to gain python's modulo behavior in Gleam, fixes for PHP guide

inoas:


inoas:
Notice, that our remainder operator behavior is expected to behave the same on all languages, including Elm, except for Python, given the current guides we have.

It will be different for say Haskell or Ruby, if we have have such a guide, see <https://torstencurdt.com/tech/posts/modulo-of-negative-numbers/>

lpil (approved):
Thank you!

PR #186, opened 2022-10-26
Fix bug in the sponsor section

lsfgrd:
The markup is broken when the sponsor has `square_avatar` set to true:
image

This PR fixes this issue by moving the `{% endunless %}` statement to inside the closing quote.


lpil (approved):
Thank you!!

PR #187, opened 2022-10-27
Changed pipe operator documentation

rikashore:
The documentation in the language tour about the pipe operator uses the function `string_builder.new` as an example, however `new` does not take a string, this changes the documentation to use `string_builder.from_string` instead.


lpil (approved):
Thank you!

PR #208, opened 2022-12-11
Updating String Elm reference using the string Gleam operator `<>`

csaltos:
Just a minor sync with the new things Gleam has now compared with Elm


lpil (approved):
Thank you

PR #214, opened 2022-12-18
Add LSP CLI reference

rockerBOO:
I understand that maybe we wouldn't want to have people use this command, but for completeness I feel it is good to have. I left the note about being used by editors.


lpil (approved):
Thank you

PR #216, opened 2022-12-21
Fix broken link to rebar3 installation steps

ducdetronquito:
Hi !

I was trying to install Gleam on my machine to toy a bit with it, and I found that the link to rebar3 installation steps does not exist anymore.

I hope the new link I provided is the correct one !

Cheers :)


lpil (approved):
Thank you!

PR #221, opened 2023-02-01
fix: code comments going out of bounds

LePichu:
Fixes `gleam.toml` code comments going out of bounds on Mobile displays.

Before: 
image

After:
image


LePichu:
Better picture of issue: 
image

LePichu:
> Thanks!
> 
> Could we actually put this in code blocks in general? There is a chance of long lines that are not comments in future too

It targets `.hljs-comment` as of now, but I can make it target `.code-comment` too.

aslilac:
fwiw, the more common way I usually see this fixed is to make the code blocks horizontally scrolling

lpil (approved):
Thank you 💜

Issue #225, opened 2023-02-03
Add links to stdlib in language tour where applicable


JohnBjrk:
Some chapters in the book about types have obvious counterparts in the stdlib. It would be nice to provide a link for diving deeper into how the stdlib can be utilised when working with these types.

For example the chapter on `Result` already have a paragraph mentioning the stdlib so it would be easy to add something like this:

>The standard library [gleam/result](https://hexdocs.pm/gleam_stdlib/gleam/result.html) module contains helpful functions for
>working with the `Result` type, make good use of them!
>
> ## Stdlib references
> * [gleam/result](https://hexdocs.pm/gleam_stdlib/gleam/result.html)

The exact format can be discussed but if would be nice if it was somewhat consistent through the book.

Some other chapters that could get a link to stdlib are:
* Bit String
* Bool
* Functions
* Ints and floats
* Lists (maybe link to iterator and set also)
* Strings

lpil:
Sounds good!

PR #226, opened 2023-02-11
📝 Added links to stdlib in tour

JohnBjrk:
Added links to stdlib according to #225


lpil (approved):
Thank you!

JohnBjrk:
Was I supposed to generate the html version and check in as well? I thought that that would be done in CI, but could not find anything like that when I checked the CI flow.

lpil:
I normally do that :)

PR #235, opened 2023-02-25
Parse CLI for the reference docs instead of writing them by hand

fabjan:
## Problem

1. The docs do not include all commands and options that are available, I presume this is because they are hand-written and maintained separately.
2. The CLI docs need to be updated by hand when the CLI is updated.

## Solution

Let's generate the docs from the CLI instead!

There is [clap-markdown] but in order to use that it looks like the gleam compiler-cli would need to upgrade clap to 4.x
([clap-md] looks too old instead). I tried cloning the compiler and bumping to clap 4.0 but there were no longer any colors in the output and enum options need rejigging to replace `possible_values` as per https://github.com/clap-rs/clap/discussions/4264.

This was more than I felt like taking on so I made a shell script for this PR instead, to reflect the CLI using its help function and render markdown from it.

It works pretty well, but is admittedly a big hack. But it is a hack that reduces toil! I have no idea how the website is built though, perhaps the `gleam` CLI is not readily available (I couldn't find any workflows to test). Given the hacky nature of the solution, and that I have no idea how this in reality slots into CI somewhere I will not be the least upset if the PR is closed because a better solution is preferred (maybe [clap-markdown]) and the status quo is good enough.

[clap-markdown]: https://github.com/ConnorGray/clap-markdown
[clap-md]: https://github.com/clap-rs/clap-md

## Effects

* Commands are documented in the order the CLI shows them (alphabetically, it seems)
* Docs for `gleam deps update` are now available!
* Options for some commands that showed none are now shown!
* Some hand-written extra details are lost :(

I purposefully made this WIP commit generating docs for the same commands in the same order as the current version, to get a somewhat readable diff on what the actual difference is: https://github.com/gleam-lang/website/commit/79e33976505b70b80ef73b53107e04d63141c10b
(note that you can see the diff in the rendered Markdown as well: https://github.com/gleam-lang/website/commit/79e33976505b70b80ef73b53107e04d63141c10b?short_path=bdd2bb7#diff-bdd2bb721d8e41a95229995f4705ccdcc79b30e74b228543bdb1708716d7c65c)

## Notes That Go Into Way Too Much Detail

The option parsing in particular is hacky, options can span multiple lines and different subcommands print in different formats for some reason:

```
❯ gleam help publish
gleam-publish 
Publish the project to the Hex package manager

This command uses this environment variables:

- HEXPM_USER: (optional) The Hex username to authenticate with.
- HEXPM_PASS: (optional) The Hex password to authenticate with.

USAGE:
    gleam publish [OPTIONS]

OPTIONS:
    -h, --help
            Print help information

        --replace
            

    -y, --yes
            
```

```
 gleam help test   
gleam-test 
Run the project tests

USAGE:
    gleam test [OPTIONS] [ARGUMENTS]...

ARGS:
    <ARGUMENTS>...    

OPTIONS:
    -h, --help                 Print help information
        --runtime <RUNTIME>    
        --target <TARGET>      The platform to target
```

```
❯ gleam help new
gleam-new 
Create a new project

USAGE:
    gleam new [OPTIONS] <PROJECT_ROOT>

ARGS:
    <PROJECT_ROOT>    Location of the project root

OPTIONS:
        --description <DESCRIPTION>    Description of the project [default: "A Gleam project"]
    -h, --help                         Print help information
        --name <NAME>                  Name of the project
        --skip-git                     Skip git initialization and creation of .gitignore, .git/*
                                       and .github/* files
        --skip-github                  Skip creation of .github/* files
        --template <TEMPLATE>          [default: lib] [possible values: lib]
```


Makefile
```diff
@@ -10,3 +10,6 @@ serve: ## Run the book dev server
 .PHONY: help
 help:
 	@cat $(MAKEFILE_LIST) | grep -E '^[a-zA-Z_-]+:.*?## .*$$' | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+writing-gleam/command-line-reference.md: writing-gleam/command-line-reference.sh
```
lpil, line 14: Can we have `make` call this automatically? 🙏

writing-gleam/command-line-reference.sh
```diff
@@ -0,0 +1,100 @@
...
+  set +e
+  # Some subcommands empty lines in the middle of the options lists, so we cannot use `sed -n '/OPTIONS:/,/^$/p'`.
+  # There are some options that span multiple lines, we need to join them.
+  sed -n '/OPTIONS:/,//p' | grep -v 'OPTIONS:' | tr '\n' '§' | sed -E 's/§ *-/\n-/g' | sed 's/§ */ /g' | trim | grep -v -- '--help'
+  set -e
+}
+
+# Render markdown help for a subcommand, or a subcommand under it.
+show_docs() {
+    subcommand="$1"
```
lpil, line 36: Could we have two space indentation in this function please?

lpil (commented):
This looks really good! I'm surprised how clear the shell script is. ✨ 

One thing I'm worried about is when we upgrade to the next version of clap, which we should, it may not work as the format has changed as you have said. I've not seen the new format, do you think it would be difficult to update to?

writing-gleam/command-line-reference.sh
fabjan: definitely , (and I should look into getting a shell formatter for my editor)

Makefile
fabjan: Sure can, the Makefile was one of the things I didn't want to touch since I don't know how/when this repo is built.

fabjan:
> One thing I'm worried about is when we upgrade to the next version of clap, which we should, it may not work as the format has changed as you have said. I've not seen the new format, do you think it would be difficult to update to?

I believe it will be quite simple as the changes are not huge:

* "SUBCOMMANDS" is now called "Commands"
* headings are no longer all caps
* the usage example is hoisted to its heading line

I updated the script to exit if it believes the clap 3 is not used, and added printing and comments to help migrate when the time comes.

This is what I get with clap 4 in a local gleam fork (not a complete update, since I did not fix the enum parsing):

```
❯ ../target/debug/gleam help publish
Publish the project to the Hex package manager

This command uses this environment variables:

- HEXPM_USER: (optional) The Hex username to authenticate with.
- HEXPM_PASS: (optional) The Hex password to authenticate with.

Usage: gleam publish [OPTIONS]

Options:
      --replace
          

  -y, --yes
          

  -h, --help
          Print help (see a summary with '-h')

```

Side note: `gleam help publish` produces an awful amount of empty lines in the "Options" section both with clap 3 and 4 in my terminal.

writing-gleam/command-line-reference.sh
lpil: If you find a good one let me know!

Makefile
lpil: Actually can we inline the calling of the script into the `build` command? I think this current one will make it so that we have to `touch writing-gleam/command-line-reference.sh` to update the page as make will think it has already been build.

Makefile
```diff
@@ -1,10 +1,13 @@
 .PHONY: build
-build: ## Build the book
+## Build the book
```
lpil, line 2: Sorry, could you put these documentation comments back, it's used by the `help` command

Makefile
fabjan: It is now inlined! I noticed it was kinda slow so debugged and cut down on the number of `gleam help` calls. If we're gonna do it every time might as well not waste time.

Makefile
fabjan: NP, I suspected it was something like that, but was apparently not observant enough to see the help rule right there! I just moved them when I added a dependency because it made the line quite long so it was a pointless change (especially now that the build is inlined).

lpil (approved):
Thank you!!!

fabjan:
Side note on the clap 4 business: once the CLI is updated, it could use https://github.com/ConnorGray/clap-markdown and shell hack can be yeeted. It will just serve as a way to reduce toil until then and was quick (and fun) enough to bash out that I felt it could be worth it.

lpil:
I mean, if this works, it works!

PR #236, opened 2023-02-26
Provide arg to mist.run_service, required for demo to compile/run.

wmealing:
The current demo doesn't compile without the third argument, this is a modification provides the third argument and allows the demo to compile.


lpil (approved):
Thank you!!

PR #239, opened 2023-03-05
Update the Gleam for Python users cheatsheet

jasperges:
When I was reading through the cheatsheet, I noticed it mentioned that Python 3.10 was in alpha at the time of writing. In the mean time Python 3.11.2 is already released. So I thought it would be helpful to update the doc. When going through it I noticed some other things that I updated.

The main things I updated:

- Updated the flow control section because Python 3.10 was already released quite some time ago.
- Updated some code examples to be more comparable. The code for Python and Gleam were sometimes a bit different. I thought it would be nice to have them describe the same things. For example for custom types Python used a `Vector` and Gleam `Headers`, now they both use `Headers`.
- Updated the Gleam `try` example to `use ... <- then(...)` following the v0.27 release.
- Tweaked some more bits and bobs

### DIFF
--- a/cheatsheets/gleam-for-python-users.md
+++ b/cheatsheets/gleam-for-python-users.md
@@ -47,8 +47,8 @@ In Python, comments are written with a `#` prefix.
 A docstring (matching """) that occurs as the first statement in a module, function, class, or method definition will become the `__doc__` attribute of that object.
 
 ````py
-def a_method():
-    """some documentation for this method"""
+def a_function():
+    """Return some important data."""
     pass
 ````
 
@@ -132,9 +132,7 @@ These annotations are accessible at runtime via the `__annotations__` module-lev
 These hints will mainly be used to inform static analysis tools like IDEs, linters...
 
 ```py
-from typing import List
-
-some_list: List[int] = [1, 2, 3]
+some_list: list[int] = [1, 2, 3]
 ```
 
 #### Gleam
@@ -182,7 +180,8 @@ mul(1, 2)
 
 #### Python
 
-In Python, top level functions are exported by default. There is no notion of private module-level functions.
+In Python, top level functions are exported by default. There is no notion of
+private module-level functions.
 
 #### Gleam
 
@@ -353,7 +352,7 @@ There is no notion of constant variables in Python.
 
 ```python
 # in the global scope
-the_answer = 42
+THE_ANSWER = 42
 ```
 
 #### Gleam
@@ -495,17 +494,14 @@ map.from_list([#("key1", "value1"), #("key2", 2)]) // Type error!
 
 ### Case
 
-Case is one of the most used control flow in Gleam. It can be seen as a switch statement on steroids. It provides a terse way to match a value type to an expression.
+Case is one of the most used control flows in Gleam. It can be seen as a switch
+statement on steroids. It provides a terse way to match a value type to an
+expression. Gleam's `case` expression is fairly similar to Python's `match`
+statement.
 
 #### Python
 
-Python does not have switch statements. `if/else` statements paired to the `type` function let the developer to write code that depends on the type and value of a variable.
-
-Until now, there has been no official comparable feature for the case expression used in Gleam.
-
-However this will change in Python 3.10 (currently in [alpha release]([https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching]), at the time of writing) that ships structural pattern matching allowing:
-
-- matching on primitive types
+Matching on primitive types:
 
 ```py
 def http_error(status):
@@ -518,7 +514,7 @@ def http_error(status):
             return "I'm a teapot"
 ```
 
-- matching on tuples with variable capturing:
+Matching on tuples with variable capturing:
 
 ```py
 match point:
@@ -534,7 +530,7 @@ match point:
         raise ValueError("Not a point")
 ```
 
-- matching on type constructors:
+Matching on type constructors:
 
 ```py
 match point:
@@ -550,7 +546,7 @@ match point:
         print("Not a point")
 ```
 
-The match expression will support guards, similar to Gleam:
+The match expression supports guards, similar to Gleam:
 
 ```py
 match point:
@@ -614,25 +610,25 @@ Python uses the notion of exceptions to interrupt the current code flow and pop 
 An exception is raised using the keyword `raise`.
 
 ```py
-def an_fn_that_fails():
-    raise Exception('an error')
+def a_function_that_fails():
+    raise Exception("an error")
 ```
 
 The callee block will be able to capture any exception raised in the block using a `try/except` set of blocks:
 
 ```py
 try:
-    print('executed')
-    an_fn_that_fails()
-    print('not_executed')
+    print("executed")
+    a_function_that_fails()
+    print("not_executed")
 except Exception as e:
-    print('doing something with the exception', e)
+    print("doing something with the exception", e)
 
 ```
 
 #### Gleam
 
-In contrast in gleam, errors are just containers with an associated value.
+In contrast in Gleam, errors are just containers with an associated value.
 
 A common container to model an operation result is `Result(ReturnType, ErrorType)`.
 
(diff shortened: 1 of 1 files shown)


cheatsheets/gleam-for-python-users.md
```diff
@@ -650,19 +652,23 @@ case parse_int("123") {
...
+for more information.)
 
-- bind a value to the providing name if Ok(Something) is matched
+- bind a value to the providing name if `Ok(Something)` is matched
 - **interrupt the flow** and return `Error(Something)`
 
 ```gleam
 let a_number = "1"
-let an_error = Error("ouch")
+let an_error = ouch"
```
aslilac, line 666: missing a `"` here

cheatsheets/gleam-for-python-users.md
```diff
@@ -614,19 +616,19 @@ Python uses the notion of exceptions to interrupt the current code flow and pop
 An exception is raised using the keyword `raise`.
 
 ```py
-def an_fn_that_fails():
-    raise Exception('an error')
+def a_fn_that_fails():
```
aslilac, line 619: "a ef en" sounds very wrong, but "an function" does as well. just depends on how you read it.

I think we should go with `a_function_that_fails` so that we don't have to argue about it. 😅

cheatsheets/gleam-for-python-users.md
```diff
@@ -614,19 +616,19 @@ Python uses the notion of exceptions to interrupt the current code flow and pop
...
 
 The callee block will be able to capture any exception raised in the block using a `try/except` set of blocks:
 
 ```py
 try:
-    print('executed')
-    an_fn_that_fails()
-    print('not_executed')
+    print("executed")
+    a_fn_that_fails()
```
aslilac, line 628: "

aslilac (commented):
deferring to @lpil since they're definitely more experienced in python, but skimmed through and had a couple thoughts

cheatsheets/gleam-for-python-users.md
jasperges: Oops, fixed.

cheatsheets/gleam-for-python-users.md
jasperges: I completely agree and updated the function name.

cheatsheets/gleam-for-python-users.md
```diff
@@ -47,9 +47,8 @@ In Python, comments are written with a `#` prefix.
 A docstring (matching """) that occurs as the first statement in a module, function, class, or method definition will become the `__doc__` attribute of that object.
 
 ````py
-def a_method():
-    """some documentation for this method"""
-    pass
+def a_function():
+    """Return some important data."""
```
lpil, line 51: Why is it that the `pass`s have been removed from the functions? To me the code looks incomplete now.

lpil (commented):
Nice, thank you! I've left one question

cheatsheets/gleam-for-python-users.md
jasperges: The reason I removed them is, that they aren't needed to be valid Python code. But from a documentation standpoint it's probably better to leave them in or use an ellipsis. If you want I can add them back, no problem.

jasperges:
One question for the MR: do you have a preference if I should make several commits if I add extra changes (what I did now) or put everything in one commit (amending it and force pushing to my branch)?

cheatsheets/gleam-for-python-users.md
lpil: Yes please! Just to make that extra bit clear for people bad at Python like myself :)

lpil:
I don't mind @jasperges , whatever you prefer 💜

cheatsheets/gleam-for-python-users.md
jasperges: Added it back. :)

jasperges:
> I don't mind @jasperges , whatever you prefer 💜

Okay, I've put it in 1 commit so you have a nice clean git history. 😇

cheatsheets/gleam-for-python-users.md
```diff
@@ -182,7 +180,10 @@ mul(1, 2)
 
 #### Python
 
-In Python, top level functions are exported by default. There is no notion of private module-level functions.
+In Python, top level functions are exported by default. There is no notion of
+private module-level functions. As a convention functions prefixed with `_` are
+considered private, but there is nothing from preventing you to use those
+outside of the module.
```
aslilac, line 186: This expanded prose kinda feels more like "Python for Gleam users" than "Gleam for Python users"

cheatsheets/gleam-for-python-users.md
```diff
@@ -499,11 +500,14 @@ Case is one of the most used control flow in Gleam. It can be seen as a switch s
...
-Until now, there has been no official comparable feature for the case expression used in Gleam.
+Before the release of Python 3.10 there hasn't been a comparable feature for
+the case expression used in Gleam. `if/else` statements paired to the `type`
+function let the developer write code that depended on the type and value of a
+variable.
 
-However this will change in Python 3.10 (currently in [alpha release]([https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching]), at the time of writing) that ships structural pattern matching allowing:
+However this changed in Python 3.10 when [structural pattern
+matching]([https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching])
+was introduced. This allows:
```
aslilac, line 510: This is a whole lot of words to say "Python 3.10 added [structural pattern matching], which is comparable to Gleam's `case` expression" 😅

cheatsheets/gleam-for-python-users.md
```diff
@@ -650,19 +654,23 @@ case parse_int("123") {
 }
 ```
 
-In order to simplify this construct, we can use the `try` keyword that will:
+In order to simplify this construct, we can use the `use` expression with the
+`then` function from the `gleam/result` module. (This used to be the `try`
+keyword. See [the v0.27 release
+post]([https://gleam.run/news/v0.27-hello-panic-goodbye-try/#moving-beyond-try])
+for more information.)
```
aslilac, line 661: I don't think we need to mention `try` anymore, it's enough to mention the blessed path

aslilac (commented):
sorry, I have one last batch of comments 😅

cheatsheets/gleam-for-python-users.md
```diff
@@ -755,6 +782,7 @@ An important difference to note is there is no OOP in Gleam. Methods can not be
 ### Unions
 
 In Python unions are declared using the `Union` type from the `typing module`
+or with the newer `|` syntax.
```
lpil, line 785: Let's remove the reference to the `Union` and only mention the current `|` syntax.

cheatsheets/gleam-for-python-users.md
```diff
@@ -765,6 +793,7 @@ two different types they must be wrapped in a new custom type.
 from typing import Union
 
 OptionalString = Union[str, None]
+OptionalString = str | None  # newer syntax
```
lpil, line 796: Let's remove the old syntax

jasperges: Removed the old syntax and also updated the Python example to be similar to the Gleam example.

cheatsheets/gleam-for-python-users.md
jasperges: That's a fair assessment. Removed the extra information. :sweat_smile:

cheatsheets/gleam-for-python-users.md
jasperges: I added a similar sentence like you suggest and moved it directly under the 'Case' header. That seemed most in line with how it is done in other sections. Now the Python and Gleam sections just show examples of how it's done in either language.

Do you think it would be helpful if the examples would be more similar or is that not really needed?

cheatsheets/gleam-for-python-users.md
jasperges: What `try`? :wink:

lpil (approved):
Wonderful! Thank you very much. And thanks for tolerating our nit-picking :)

jasperges:
> Wonderful! Thank you very much. And thanks for tolerating our nit-picking :)

You are more then welcome! It was good nit-picking that improved the documentation. :)

Thank you for the wonderful language! I'm still a Gleam novice (did a bunch of exercises on Exercism), but so far I quite like it.

PR #240, opened 2023-03-06
Fix minor typo in javascript filename.

wmealing:
The javascript filename presented ends in .mjs, which is probably a typo.   I am new to javascript, so perhaps it is not.  

Thank you for considering this MR.


inoas:
https://nodejs.org/docs/latest/api/esm.html#esm_differences_between_es_modules_and_commonjs
node treats .mjs and .cjs different from .js it seems

wmealing:
Ah righto, i'll close this one up then.

PR #251, opened 2023-05-22
Replace `assert` with `let assert`

aleksrutins:
I was reading through the docs, and realized that the "Gleam for <x> users" pages still used `assert` instead of `let assert`. This PR fixes that.


lpil (approved):
Thank you!

PR #253, opened 2023-05-27
Use Gleam package from nixpkgs.

vic:
Now that NixOS/nixpkgs repository contains a community supported [derivation ](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/gleam/default.nix) for Gleam, it is better to point users to use that package instead of using vic's gleam-nix flake.


lpil (approved):
Oh cool! Thank you!!

PR #254, opened 2023-05-29
add directions for using external types

gordonguthrie:
Raised a documentation change to explain how to import external types correctly


lpil:
Thank you, but I think this is already covered in the module section: https://gleam.run/book/tour/modules.html

Perhaps that could be improved in some fashion.

PR #271, opened 2023-08-16
Document gleam_otp usage

gamecubate:
[The website](https://gleam.run/)'s `Reliable and scalable` section example is missing some code to make it beginner-friendly. The code in this PR adds these missing lines.

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,20 +51,31 @@ content=header_content %}
         scale and stay lightning fast with ease.
       </p>
     </div>
-    <pre><code><span class="code-keyword">fn</span> <span class="code-fn">spawn_task</span>(i) {
-  <span class="code-module">task.</span><span class="code-fn">async</span>(<span class="code-keyword">fn</span>() {
-    <span class="code-keyword">let</span> n = <span class="code-module">int.</span><span class="code-fn">to_string</span>(i)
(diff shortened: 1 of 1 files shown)


gamecubate:
I appreciate that.

For my part, I came into gleam a few months ago, hoping to find an easier-to-read syntax than Erlang's -- check -- and less noise than Elixir -- check -- for my OTP app needs. The home page code snippets hinted at that.

However, I quickly realized that the spawn_task segment was missing imports and finding the answer in the documentation took longer than the time I add for that. I decided to put my evaluation aside and revisit the project a few months later.

3 months ago, same results; same postponement decision.

Today, seeing as nothing had changed, I decided to dig a bit, found the answer, and submitted this PR, knowing that it would help others in the same position (not much time, check out these languages; pick one now) and might actually increase traction for the project if that is a consideration of yours.

Food for thought. I will continue nonetheless my evaluation of Gleam as I like what I've seen.

lpil:
While we want more information for newcomers, I think adding it here was obscure what it is trying to communicate, a brief taste of the language and its main features.

I had a look at other languages, and I couldn't find any that offer this level of context on their homepage. Elixir, for example, offers less and doesn't show that the code has to be put in a function, let alone the module and how to create the project.

Issue #200, opened 2022-11-24
Custom Types chapter has a section on Erlang but not on JavaScript InterOp


inoas:
https://gleam.run/book/tour/custom-types.html#erlang-interop

(... or Elixir or TypeScript)

lpil:
Good idea!

ankddev:
@lpil Looks like, it's fixed and could be closed!

PR #272, opened 2023-09-09
Add section on JavaScript interop for custom types

nino:
For #200.

Does something like this look reasonable?

### DIFF
--- a/book-src/tour/custom-types.md
+++ b/book-src/tour/custom-types.md
@@ -266,9 +266,9 @@ Cat(name: "Jinx", cuteness: 2002)
 
 ```javascript
 // JavaScript
-import { Guest, LoggedIn, Animal } from "my_module.mjs";
+import { Guest, LoggedIn, Cat } from "my_module.mjs";
 
 new Guest();
 new LoggedIn("Kim");
-new Animal("Jinx", 2002);
+new Cat("Jinx", 2002);
 ```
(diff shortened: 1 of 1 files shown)


book-src/tour/custom-types.md
```diff
@@ -250,3 +250,23 @@ LoggedIn("Kim")
...
+Guest
+LoggedIn("Kim")
+Cat(name: "Jinx", cuteness: 2002)
+```
+
+```javascript
+// JavaScript
+new Guest();              // => {}
+new LoggedIn("Kim");      // => { 0: "Kim" }
+new Animal("Jinx", 2002); // => { name: "Jinx", cuteness: 2002 }
```
lpil, line 271: Could you remove these comments? They're not accurate, and I think the constructors on the left are clear enough.

lpil (commented):
Nice! Do you think we should also show them being imported?

book-src/tour/custom-types.md
nino: Yeah, I was trying to find a way to explain how to access named and unnamed properties. Maybe we could do something like

```javascript
const user = new LoggedIn("Kim");
user[0] === "Kim";

const cat = new Cat("Jinx", 2002);
cat.cuteness === 2002;
```

Not sure if that's overkill

nino:
> Do you think we should also show them being imported?

That's a good idea actually

lpil (approved):
Wonderful! Thank you

PR #274, opened 2023-09-17
Add JavaScript cheatsheet

nino:
https://github.com/gleam-lang/gleam/issues/945

I started adapting the Elixir cheatsheet for JavaScript/TypeScript.

Left to do:

- Better subtitle
- Table of contents
- The rest of the doc

Hope it's ok to put the draft PR here; it not, I can close it until I'm fully done :)


cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+ */
+class Bar {}
+
+/**
+ * A quux function.
+ *
+ * @param {string} str    String passed to quux
+ * @returns {string}      An unprocessed string
+ */
+function quux(string) { return str; }
```
lpil, line 63: Could you remove all the foobar references please

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+function sum(x, y) {
+  return x + y;
+}
+
+const mul = (x, y) => x * y;
+```
+
+#### Gleam
+
+Gleam's functions are declared using a syntax similar to Rust. Gleam's anonymous
```
lpil, line 161: Remove the Rust reference please

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+functions have a similar syntax, just without the function name.
+
+```gleam
+pub fn sum(x, y) {
+  x + y
+}
+
+let mul = fn(x, y) { x * y }
+mul(1, 2)
```
lpil, line 170: This code implies you can have top level expressions as if Gleam were a scripting language, but this is not the case

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+Gleam has a special function-capturing syntax:
+
+```gleam
+fn add(a: Int, b: Int) -> Int {
+  a + b
+}
+
+let add2 = fn(x) { add(x, 2) }
+// is equivalent to:
+let add2 = add(_, 2)
```
lpil, line 185: Here too

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+pub fn add(x: Int, y: Int) -> Int {
+  x + y
+}
+
+pub fn mul(x: Int, y: Int) -> Bool { // compile error, type mismatch
+  x * y
+}
+```
+
+### Referencing functions
```
lpil, line 258: I think we can remove this section as JS programmers wouldn't expect anything else

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+import my_library/tools/math
+
+let pixel = math.Green
+let sum = math.add(2, 2)
+
+// Or specify exactly what you want
+import my_library/tools/math.{type Color, Red, add}
+
+let colors: Color = [Red, math.Green]
+let sum = add(2, 2)
```
lpil, line 383: This looks like scripting

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+| Divide            | `/`        | `/`   | In Gleam both values **ints**                  |
+| Divide            | `/`        | `/.`  | In Gleam both values must be **floats**        |
+| Remainder         | `%`        | `%`   | In Gleam both values must be **ints**          |
+| Concatenate       | `+`        | `<>`  | In Gleam both values must be **strings**       |
+| Pipe              |            | `|>`  | Gleam's pipe can pipe into anonymous functions |
+
+### Notes on operators
+
+- JavaScript operators are short-circuiting as in Gleam.
+- Gleam's `/` operator always returns an integer.
```
lpil, line 418:
```suggestion
- Gleam's `/` operator always returns an int.
```

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+`string_builder` via its standard library for performant string building.
+
+```gleam
+"Hellø, world!"
+```
+
+### Tuples
+
+Tuples are very useful in Gleam as they're the only collection data type that
+allows mixed types in the collection.
```
lpil, line 547: Custom types permit this too

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+```javascript
+let list = [2, 3, 4];
+list = [1, ...list, 3];
+const [firstElement, secondElement, ...rest] = list;
+console.log(["hello", ...list]); // works
+```
+
+#### Gleam
+
+Gleam has a "cons" operator that works for lists destructuring and pattern
```
lpil, line 590: We don't have a cons operator, we have a list-prepend syntax

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+
+(You would usually translate a TypeScript `type` or `class` to a Gleam custom
+type, and only use a Gleam `Map` for arbitrary key–value pairs, equivalent to a
+TypeScript `Map` or `Record`.)
+
+```gleam
+import gleam/dict
+
+dict.from_list([#("key1", "value1"), #("key2", "value2")])
+dict.from_list([#("key1", "value1"), #("key2", 2)]) // Type error!
```
lpil, line 662: This looks like scripting

cheatsheets/gleam-for-javascript-users.md
```diff
@@ -0,0 +1,700 @@
...
+
+To make this more readable, Gleam has the pipe operator `|>`
+
+```gleam
+import gleam/string
+
+"hello, world"
+|> string.uppercase
+|> string.repeat(2)      // defaults to piping into the first argument
+|> string.split(_, ",")  // you can use _ to specify the argument to pipe into
```
lpil, line 699: Would be good to pipe into the second position as the first is the default

lpil (commented):
Thank you! Sorry for taking so long, you've been at the end of the inbox for far too long and I'm only just getting through all the notifications

cheatsheets/gleam-for-javascript-users.md
josealonso: Although its on the gleam tour already, I think it's worth repeating there is no **explicit return**, because many JS developers might not know other functional language.

PR #275, opened 2023-09-21
Larger font and list spacing

mewfree:
Following this post by @lpil: https://twitter.com/louispilfold/status/1704624186830807255
I visited other pages and the biggest difference from the homepage was the smaller font (18px vs 22px on the homepage) and lists feeling a bit "squished"

This keep font size consistent between homepage and other pages at 22px and adds a little bit of margin under each list element 

# Before
Screenshot 2023-09-20 at 21 36 49

# After 
Screenshot 2023-09-20 at 21 36 32

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -55,7 +55,7 @@
   /* TODO: condense these into a few appopriatedly named variables */
   --font-size-small: 12px;
   --font-size-code: 16px;
-  --font-size-normal: 18px;
+  --font-size-normal: 22px;
   --font-size-navbar: 18px;
   --font-size-hero-button: 24px;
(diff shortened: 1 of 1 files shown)


lpil:
Hello! Thanks for this, but the aim is not to make the other pages match the homepage more closely, but instead to make it a comfortable and clear reading experience. This font size is uncomfortably large to read, and larger than any of the other websites I checked as a reference.

aslilac:
I think the additional list spacing is nice, but yeah the font is way too big

mewfree:
No problem, hope I can be more helpful in the future!

lpil:
Thank you 💜

PR #281, opened 2023-10-13
Move language tour from mdbook to single page with gleam website style

pietroppeter:
This moves the language tour: https://gleam.run/book/tour/ from mdbook to a page in the standard gleam website.
The rationale for this is that gleam website style is really nice and I do feel a bit bad when I change to mdbook when using language tour. Mdbook itself is an awesome tool, the motivation for this is more about consistency of the Gleam branding.

Here are a couple of screenshot on how it would look:

Screenshot 2023-10-13 alle 17 59 43

Screenshot 2023-10-13 alle 18 00 03

By moving to a single page we lose some features provided by mdbook, such as search (you can revert to standard search in the single page though) and a nice toc that is always present on the side. A toc has been added to link with main headings and the back browser button can be used to go back to the toc. In principle a nicer floating toc could be added later (for someone practical of web design - or someone more brave than me to use chatgpt or related - should not be too difficult, it kind of is at the moment for me).

Another advantage is that now the site is built only with Jekyll (the build using mdbook is not great since the content is currently committed in the git folder), so it is in principle simpler to build and contribute (see below on how the PR was done).

The work done here is pretty straightforward:
- a new `tour` folder has been created with an `index.md` that contains the complete tour page through `include_relative` mechanism (as usual found about it through StackOverflow: https://stackoverflow.com/a/41966993/4178189)
- all pages from `book-src` have been copied into `tour` folder; `README.md` has been converted into a `intro.md` and  `SUMMARY.md` has been changed into a `toc.md`
- the links from toc to heading is done with this `_includes\linkedHeadings.html` trick (explained [here on dev.to](https://dev.to/aleksandrhovhannisyan/create-heading-links-in-jekyll-without-any-javascript-using-includes-328h)). I was a bit surprised this is not out-of-the-box in Jekyll, probably there is an extension that adds this feature (create an anchor for each heading). This makes the main heading of each file not standard markdown (not anymore `# Tuples`), which is not great but it is not too bad either and it is a very simple solution. It also adds a big "#" to the linkable headings (which can be changed).
- very few changes were done to the text:
  - added a subtitle: `A quick tour of Gleam features` (could be improved)
  - changed `in this book` to `in this page` in `intro.md`
  - titles of "chapters" are now consistent with headings. Most of the time the change is in the heading (e.g. `# Bool` -> `# Bools`, ...) but not always (`Use` in toc becomes now `Use expression`)
  - a small fix in `external-functions.md`: a triple tick was missing at the end 
- all pages from book now (should) redirect to the single `tour/index.html` page (I also ported the redirects mentioned in `book.toml` (now removed). (I am not sure how to test the redirect, in my gitpod setup I was not able to test them).

As possible improvements that I think could be in scope for the PR are:
- a better subtitle than `A quick tour of Gleam features` or a better initial sentence than `In this page` or similar small text improvements?
- a better way to have linked headings? it would be nice to have them for the whole site, every heading
- something I do not like too much is how backticked content is rendered (e.g. the `True` and `False` in the Bool preview above), I think this could probably be improved with some css (it would likely apply to the whole site).

As an aside I did make this PR taking advantage of gitpod and the process is like this:
- take the url of the repo (of your fork if you intend to contribute) and change `www.github.com` with `gitpod.io/#github.com`
- this will guide you through creating a vs code environment where you have already ruby installed (I had trouble installing ruby on my Mac M1, this is why I used gitpod)
- the environment will already have ruby installed, you just need to `gem install github-pages webrick` and `make serve` as reported in the `README.md` and inside the browser you will have a live preview of the site with hot reload on changes
- in case you want to push your changes to a branch you will need to give gitpod the permission to write to public_repo

I think the above is a very quick and convenient way to contribute to the website repo (it is possible a similar simple workflow can be achieved with GitHub Codespaces), the main advantage is that it does not require to have a working local environment with ruby and Jekyll.

This has been briefly discussed today in #general channel on discord. There was some interest and some debate, not a complete consensus, still feel the effort for the PR is worth it even if it is decided not to proceed with this. Thanks a lot for the great language and community!


lpil:
Hello! I'm hoping to get time to review this soon. Thanks for waiting 💜

pietroppeter:
Hi, no problem, take all the time you need. Have a great day! 💜

tour/assert.md
```diff
@@ -1,4 +1,4 @@
-# Assert
+{% include linkedHeading.html heading="Assert" level=1 %}
```
lpil, line 1: Could you lowercase this file name please? We don't use any uppercase letters.

writing-gleam/command-line-reference.md
```diff
@@ -1,267 +0,0 @@
----
```
lpil, line 1: Was this file intentionally removed?

lpil (commented):
Thank you for this!!

I think the table of contents would need to be at the side of the page and stay visible always, so long as the page is wide enough. With it at the top it's not usable enough compared to the previous design. On mobile I think this would need also be accessible, but I don't know what the design should be.

I found it hard to see where one section ended and another began. It needs to be visually clear where each begins and ends.

pietroppeter:
Thanks for the review! I will address the specific comments and try to think about how to work on the table of contents to be on the side/accessible on smaller screens. Not sure when I will get to that though, very busy this week and the next.

writing-gleam/command-line-reference.md
pietroppeter: nope, no idea why it is removed, I will restore it

pietroppeter:
marking this as draft so you don't need to bother looking at it until it is ready for another review (and I will updraft it then)

pietroppeter:
I think with the new excellent language tour 🥳 this is not anymore needed and can be closed :)

lpil:
Oh yes! Thank you

PR #282, opened 2023-10-19
fix look of the unknown field error in the homepage

giacomocavalieri:
When [this PR](https://github.com/gleam-lang/gleam/pull/2375) gets merged and released in a new version of the language, this will fix the look of the error in Gleam's homepage to make it consistent with the new compiler's behaviour


lpil (approved):
Thank you

PR #285, opened 2023-11-04
Update book, expression blocks as parentheses

matt-savvy:
The phrase "can be used instead of parentheses" is ambiguous. It makes it seem like you can use parentheses or expression blocks, when in fact you can only use expression blocks.


lpil (approved):
Thank you.

PR #288, opened 2023-11-09
Add documentation for a few extra installation methods

aslilac:
- Bring Chocolatey back, since it's been receiving updates much more quickly
- New ~~and improved~~ `curl | sh` installer!

### DIFF
--- a/getting-started/installing.md
+++ b/getting-started/installing.md
@@ -31,7 +31,7 @@ The easiest way to install Gleam on Linux, Windows, and Apple macOS is to downlo
 prebuilt version of the compiler from the [GitHub release
 page](https://github.com/gleam-lang/gleam/releases).
 
-If you prefer, a script is also available to use.
+If you prefer, a script is also available to use on Linux and macOS.
 
 ```sh
 curl -fsSL https://gleam.pink/install.sh | sh
--- a/getting-started/installing.md
+++ b/getting-started/installing.md
@@ -100,6 +100,14 @@ sudo xbps-install gleam
 
 ### Windows
 
+#### Using Chocolatey
+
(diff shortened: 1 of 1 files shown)


getting-started/installing.md
```diff
@@ -31,6 +31,12 @@ The easiest way to install Gleam on Linux, Windows, and Apple macOS is to downlo
 prebuilt version of the compiler from the [GitHub release
 page](https://github.com/gleam-lang/gleam/releases).
 
+If you prefer, a script is also available to use on Linux and macOS.
+
+```sh
+curl -fsSL https://gleam.pink/install.sh | sh
```
lpil, line 37: This looks official but it's not, and piping into `sh` is very dangerous, so I'm unsure about this addition.

If we do have this I think the script should confirm with the user before making changes (it modifies the system by default), and provide uninstall functionality too.

What do you think about having it install Erlang too? I would expect an installer to install everything that is required by default, like the OS packages do.

aslilac: imo, `curl | sh` isn't any more dangerous than running an executable you downloaded from the internet. it's not an ideal installation method, but it's super practical, and not really _that_ much scarier.

I don't care to add a confirmation or provide an uninstaller. it already creates a Gleam specific prefix directory by default, and uninstall is just deleting that directory; but it allows choosing a custom prefix! the uninstall script would have no reliable way to know if you used the default, or specified your own, and where that would be.

I might add support for installing Erlang at some point, but the real problem I was trying to solve is that a lot of distros don't have a Gleam package. they _do_ mostly have Erlang packages tho.

aslilac: also from a quick skim, I can't find a good source for prebuilt Erlang, and I've been avoiding treading into package manager territory 💀

PR #289, opened 2023-11-13
Update gleam book type import examples

matt-savvy:
## Changes

- update book order, move custom types before modules. 
The section on modules references custom types, so ideally the reader should know what those are first. 
- update modules section on unqualified imports, add exampling showing use of `type` keyword to import a type vs a constructor.


lpil (approved):
Thank you

PR #296, opened 2024-01-18
Fix invalid external function directive

wingyplus:
Since the `JavaScript external functions`, it'll get compilation error because `@external` missing `,` between argument 1 (module, file) and 2 (function). This PR fix this issue.


lpil (approved):
Thank you

PR #301, opened 2024-02-19
Writing gleam

lpil:


icp1994:
I think this broke a couple links in https://gleam.run/documentation/.
* The `Writing Gleam` bullet doesn't have a link
* The `Getting started` goes to `https://gleam.run//writing-gleam/` (maybe that's intentional , but weird two `/`)

lpil:
Thank you

PR #305, opened 2024-02-24
Add link link to Exercism in documentation

human154:


lpil (commented):
This isn't documentation so I don't think this is the right place to put this. Or at least not in the first main section there.

Perhaps we could have another section lower for external content and courses?

human154:
...again, my goal is just to make the exorcism content a bit more discoverable. i.e. more so than hoping people will find it many lines down on the "awesome gleam" list.  This way all that community learning content is linked to directly and becomes more discoverable.

human154:
How does that look?

lpil (approved):
Thank you

PR #307, opened 2024-02-26
Update documentation page: Deploying on Fly.io

pintomena:
- Update Dockerfile section code to the last image
- Fix micro web app code sample: update module from `bit_builder` (removed in stdlib v0.35.0) to `byte_builder`


lpil (approved):
Thank you

PR #313, opened 2024-03-04
fix: example code importing a package which doesn't exist

Joshswooft:
Congrats on the V1 release, I was eager to try about gleam and noticed a small issue on one of the pages.

The package name is incorrect, see here: https://hexdocs.pm/gleam_stdlib/gleam/bytes_builder.html#BytesBuilder


lpil (approved):
Thank you!

PR #316, opened 2024-03-05
fix: convert strokes into paths

ffigiel:
People at exercism ran into some issues with working with Lucy images in Figma. I applied the fixes suggested in the linked video: https://discord.com/channels/768594524158427167/768594524158427170/1214646878952628295

 - use viewbox instead of width/height
 - convert strokes to shapes


ffigiel:
The github preview shows the images were scaled down, even though their internal dimensions are still the same. I suspect it's due to the viewbox change

PR #319, opened 2024-03-07
Fix missing images on the branding website

ffigiel:
After replacing dimensions with `viewBox` attributes in https://github.com/gleam-lang/website/pull/316 the images disappeared as they didn't have an explicit size


lpil (approved):
Thank you

PR #321, opened 2024-03-08
Create gleam-for-go-users guide

gamebox:
This is a first draft and needs some editing and fact checking.


cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+```gleam
+type Person {
+  Person(name: String, age: Int)
+}
+
+let person = Person(name: "Joe", age: 40)
+let name = person.name
+```
+
+An important difference to note is there is no Java-style object-orientation in
```
lpil, line 1045:
```suggestion
An important difference to note is there is no object-orientation in
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+    False -> AFloat(1.0)
+  }
+}
+```
+
+### Opaque custom types
+
+In Go, you can choose to not export a type and then they will be considered
+opaque to the user of your package.  Just have the type have a lowercase
+identifier.
```
lpil, line 1126: Can you still reference the type from other packages? I thought it would make it private, not opaque

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+
+## Architecture
+
+To iterate a few foundational differences:
+
+1. Programming model: imperative VS functional immutable
+   programming
+2. Guarantees: implicit nillability vs explicit nillability
+3. Runtime model: lightweight coroutines(`goroutines`) VS Erlang/OTP processes
+4. Error handling: error value (and very few panics) VS result type
```
lpil, line 1376: I think Go and Gleam use panics in a similar way- for fatal unexpected errors. I could be wrong though, my Go is rusty.

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+5. Distribution: Single static binaries vs Dynamically linked binary that relies on Erlang runtime
```
lpil, line 1377:
```suggestion
5. Distribution: Single static binaries vs BEAM virtual machine bytecode
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+  library.
+
+### Guarantees and types
+
+- Both Go and Gleam features strong static typing.
+- Both Go and Gleam values are not automatically cast.
+- Both Go and Gleam's comparison operators are very strict and limited, any other
+  comparisons and conversions must happen via function calls or other explicit type
+  conversion mechanisms.
+- Go's checks happen at compile time (except for reflection), Gleam uses the compiler
```
lpil, line 1408: This sounds a bit like Gleam does runtime checks while Go do more compile time checks?

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+  with Erlang, a dynamically language that relies on good interfaces written in Gleam
+  to provide any type-safety.
+- Go has limited type inference, all functions and data structures require explicit types.
+  Gleam's type inference allows you to be lazy for almost all type definitions.
+  Gleam's type system will always assist you in what types are expected and/or
+  conflicting. Gleam's type system will help you discover APIs.
+
+### Runtime model
+
+- Gleam can run on Erlang/BEAM, on the browser but also Deno or NodeJS.
```
lpil, line 1419:
```suggestion
- Gleam can run on Erlang or JavaScript runtimes.
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+- Gleam on Erlang/BEAM allows to processes requests in complete isolation,
```
lpil, line 1420:
```suggestion
- Gleam on Erlangallows to processes requests in complete isolation,
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+  unlike Go, where memory sharing is allowed - though communicating through
+  channels is recommended.  Channels and Goroutines allow for good isolation
+  if used in a disciplined manner.  That being said, Mutexes and other shared
+  state mechanisms are included in the standard library.
+
+  The level of isolation means that, if a process crashes then the supervision
+  system can restart that process or after a while or amount of tries abort
+  repeating restarts on the process with that given input data. This means
+  Erlang/BEAM will yield incredibly robust concurrent systems, even superior
```
lpil, line 1429:
```suggestion
  Erlang can yield incredibly robust concurrent systems, even superior
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+  to what experienced Gophers can accomplish with Goroutines, Channels, and
+  recover patterns.
+- When executing Gleam code in fact its compiled Erlang or JavaScript is
+  executed. So in case there are runtime crashes, the crash log will show
+  Erlang (or browser-console/NodeJS/Deno) debug information. In Gleam
```
lpil, line 1434:
```suggestion
  Erlang or JavaScript debug information. In Gleam
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+  to read, in Go applications runtime errors might be more common but the will be
+  easier to read.
+
+### Error handling
+
+- Gleam will catch all errors that are expected to happen via the `Result`
+  type. There can however be other errors, such as miss-behavior due
+  accidental to division by 0, crashes on RAM or storage limits, hardware
+  failures, etc. In these cases on the BEAM there are ways to manage these
+  via BEAM's supervision trees.
```
lpil, line 1445:
```suggestion
  via Erlang/OTP's supervision trees.
```

lpil (commented):
Very nice! This must have taken some time to write! Fantastic work.

I've left a bunch of small notes inline 🙏 

Thank you

nii236:
@gamebox are you planning on updating this PR?

gamebox:
Yes, sorry I missed Louis's comments.  I'll update sometime this week.

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+type Num interface {
+	int | float
+}
+```
+
+Unfortunately, these unions provide little value to the Go Developer due to their
+constraints.
+
+#### Gleam
+
```
nii236, line 1107: Worth having a sentence here explaining Gleam's approach

cheatsheets/gleam-for-go-users.md
nii236: It will be opaque from outside the package, which is the same thing as private? I might be misunderstanding

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+Modules in Gleam are the principal unit of code organization. Each
+Gleam file is a Module.  A module exports a number of symbols and
+can import symbols from other modules.
+
+### Go
+
+In Go, the term _modules_ refers to a set of packages that share
+common dependencies and are stored alongside one another.  A _package_
+in Go is closer to the Gleam concept of _modules_.  The biggest difference
+is that a _package_ in Go can be spread across many files in one directory.
```
nii236, line 1191: Very helpful comment! This is what tripped me up

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+Importing common types such as `gleam/order.{Lt, Eq, Gt}` or
+`gleam/option.{Some,None}` can be very helpful.
+
+## Architecture
+
+To iterate a few foundational differences:
+
+1. Programming model: imperative VS functional immutable
+   programming
+2. Guarantees: implicit nillability vs explicit nillability
```
nii236, line 1374: nillability == nullability? Or is it something else

cheatsheets/gleam-for-go-users.md
nii236: In general, panics should never be used in Go (especially in libraries), you'd return wrapped errors to build a manual one-line stack trace and handle it at the top.

Only time I've used panic is because I was developing quickly and planned to replace it all with error returns

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+  Go sometimes directly mutates references of simple values such as when using
+  pointers or global variables, though this is very much not idiomatic.
+- Gleam allows to rebind variables freely to make it easy to update data
+  structures by making a copy and binding it to the existing variable.
+- Go features a massive, powerful standard library centered around simple interfaces.
+  Everything from string manipulation to the tools to create an http server are included
+  out of the box.  It has been tuned for over 15 years for performance and has strong
+  guarantees about backwards and forwards compatibility.
+- Gleam allows you to opt into a smaller, well polished and consistent standard
+  library.
```
nii236, line 1399: I'd re-order this so that points starting with "Go" are first and points starting with "Gleam" are second. Right now the ordering is a bit harder to follow because it keeps switching between the two languages

cheatsheets/gleam-for-go-users.md
gamebox: That's true, it's rare that you will write a panic, panics will occur, such as when you read from a nil channel

gamebox: Actually I think @lpil is more or less right, a panic can be recovered and is a way for the application to choose to recover and get back into a working state.  Which is similar to having a process being killed in a supervision tree.

cheatsheets/gleam-for-go-users.md
gamebox: Same. Use nillability here since Go uses nil as opposed to null

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+let mul = fn(x, y) { x * y }
+mul(1, 2)
+```
+
+A difference between Go's and Gleam's anonymous functions is that in Go they
+create a new local scope but inherit the surrounding scope, in Gleam they close 
+over the local scope, aka create a copy and inherit all variables in the scope.
+This means that in Gleam you can shadow local variables within anonymous functions
+but you cannot influence the variable bindings in the outer scope.  In Go you can
+actually change the values in the outer scope, even if that scope has been returned.
```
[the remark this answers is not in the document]
gamebox: Right, SSA just like Erlang/Elixir

cheatsheets/gleam-for-go-users.md
nii236: Either that or have subheadings, one for gleam, one for go

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+```
+
+Slices have some unique concerns as regards to references as if you append to
+a slice, the underlying array may no longer be in the same memory location as
+it was before and some mutative operations can have unexpected behavior for the
+unpracticed (or even experienced) Gopher.
+
+#### Gleam
+
+Gleam has a `cons` operator that works for lists destructuring and
```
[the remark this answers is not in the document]
gamebox: I have no idea where I got that from, will fix

cheatsheets/gleam-for-go-users.md
gamebox: Sorry, I should clear this up.  To make a type Opaque in Go is having the type public, but all fields and methods being private.  It would then be unusable outside of whatever public functions that work with it are exposed.  The water is muddied by the fact that methods are just functions that have a certain type as the receiver.  But I think not exposing them is closer to the intent of opaque types in that they may flow through code outside the package, but that code should have no idea about the details of the type.

gamebox:
This should be ready for your review when you get a chance @lpil

gamebox:
I'd really love to close this out, if you get a chance @lpil .

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1452 @@
...
+/*
+ * Hello, Joe!
+ */
+```
+
+In Go, top level declarations can be annotated with a comment directly above it to create documentation for that symbol.
+
+```Go
+// This is a special interface
+type Foo interface {}
```
lpil, line 62: Could you swap "foobar" for something that isn't a horrors-of-war reference please 🙏

lpil (commented):
Hello! Sorry for the delay.

It seems that not all my comments have been addressed here. Let me know when the PR is ready

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1452 @@
...
+### Go
+
+In Go you specify values by:
+- Specifically declare the variable with the `var` keyboard and a name and then either just the type, or an optional type and value
+- a variable name (or comma separated list of names), the `:=` operator and then an expression
+
+```Go
+size := 50
+size := size + 100
+size = 1
```
adamdecaf, line 118: The second line is not valid following the first. You get an `no new variables on left side of :=` error. 

https://go.dev/play/p/UGW6HduxifQ

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1452 @@
...
+In Go, you can define functions with the `func` keyword. The `return`
+keyword is required if the function has a return value, and optional otherwise.
+
+```Go
+func hello(name string) string {
+  if name == 'Joe' {
+    return 'Welcome back, Joe!'
+  }
+  return "Hello $name"
+}
```
adamdecaf, line 204: You need double quotes for strings. Single quotes are runes (characters). Go doesn't have interpolation inside of strings. 

```suggestion
func hello(name string) string {
	if name == "Joe" {
		return "Welcome back, Joe!"
	}
	return fmt.Sprintf("Hello %s", name)
}
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1452 @@
...
+  // No return value
+}
+```
+
+Anonymous functions can also be defined and be bound to variables.
+
+```Go
+x := 2
+GoAnonFn := func(y) { return x * y } // Captures x
+GoAnonFn(2, 3) // 6
```
adamdecaf, line 216:
```suggestion
x := 2
GoAnonFn := func(y int) int { return x * y } // Captures x
GoAnonFn(3) // 6
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1452 @@
...
+  }
+  return "Hello $name"
+}
+
+func privateHello(name string) string { // Package private
+  if name == 'Joe' {
+    return 'Welcome back, Joe!'
+  }
+  return "Hello $name"
+}
```
adamdecaf, line 270:
```suggestion
func PublicHello(name string) string {  // Exported
  if name == "Joe" {
    return "Welcome back, Joe!"
  }
  return fmt.Sprintf("Hello %s", name)
}

func privateHello(name string) string { // Package private
  if name == "Joe" {
    return "Welcome back, Joe!"
  }
  return fmt.Sprintf("Hello %s", name)
}
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1452 @@
...
+
+In Gleam all strings are unicode binaries. Gleam however offers a `StringBuilder` type via its standard
+library which may be more performant than joining strings in some scenarios.
+
+#### Go
+
+```Go
+what := "world"
+"Hellø, world!"
+fmt.Sprintf("Hellø, %d!", what);
```
adamdecaf, line 563:
```suggestion
what := "world"
// "Hellø, world!"
fmt.Printf("Hellø, %v!", what)
```

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1452 @@
...
+```
+
+We can now call Identity from the main package in your binary.
+
+```Go
+package main
+
+import (
+  "fmt"
+m
```
adamdecaf, line 1220:
```suggestion
```

MiniAppleTheApple:
isn't it better to use the equivalent/similar example in the Unions section?

gamebox:
@MiniAppleTheApple Well, there really _isn't_ a similar approach unless you mean a struct that has a tag and then contains pointers to structs for all possible "variants" or contains all of the fields relevant to all variants.  But I personally think it's more honest to say "This really isn't a thing you would do in Go".  Plus, union semantics are just different then either of the two options I described above.

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+    case 400:
+      return "Bad request"
+    case 404:
+      return "Not found"
+    case 418:
+      return "I'm a teapot"
+    default:
+      return "Internal Server Error"
+  }
+}
```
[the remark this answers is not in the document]
gamebox: Done

cheatsheets/gleam-for-go-users.md
gamebox: Done

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+
+### Strings
+
+In Go strings are slices (dynamically sized arrays) of bytes. Those bytes are
+arbitrary, and are not required to contain Unicode or any other string encoding.
+Go's standard library however has extensive support for string operations. Strings
+can be built efficiently with a `StringBuilder` type from the `strings` package in
+the standard library.  Interpolation can be accomplished using `Sprintf` and other
+similar functions found in `fmt` package amongst others in the standard library.
+This uses C `strfmt` style format strings that are statically checked at compile time.
```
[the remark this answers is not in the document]
gamebox: Shortened it to focus on the important semantics

cheatsheets/gleam-for-go-users.md
gamebox: Done

cheatsheets/gleam-for-go-users.md
gamebox: Fixed

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+```gleam
+let list = [2, 3, 4]
+let list = [1, ..list]
+let [1, second_element, ..] = list
+[1.0, ..list] // compile error, type mismatch
+```
+
+### Maps
+
+Maps are similar in Go and Gleam, except that they are mutated in Go and
```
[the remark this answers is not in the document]
gamebox: Fixed

cheatsheets/gleam-for-go-users.md
gamebox: I don't understand?

gamebox:
All conversations are resolved, and suggestions merged.  Looks like now this should be good for a final review.

cheatsheets/gleam-for-go-users.md
adamdecaf: The m is causing a syntax error

MiniAppleTheApple:
> @MiniAppleTheApple Well, there really _isn't_ a similar approach unless you mean a struct that has a tag and then contains pointers to structs for all possible "variants" or contains all of the fields relevant to all variants. But I personally think it's more honest to say "This really isn't a thing you would do in Go". Plus, union semantics are just different then either of the two options I described above.

Even though there isn't a 100% similar way, you can still approximate both example code as close as possible. I will give a suggestion later.

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1447 @@
...
+- [Type aliases](#type-aliases)
+- [Custom types](#custom-types)
+  - [Records](#records)
+  - [Unions](#unions) IN PROGRESS
+  - [Opaque custom types](#opaque-custom-types) IN PROGRESS
+- [Modules](#modules) IN PROGRESS
+  - [Imports](#imports) IN PROGRESS
+  - [Named imports](#named-imports) IN PROGRESS
+  - [Unqualified imports](#unqualified-imports) IN PROGRESS
+- [Architecture](#architecture) IN PROGRESS
```
nii236, line 38: Are these actually "in progress"? Or should this note be removed?

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1447 @@
...
+  structures. This allows processes that fail to simply restart as there are no
+  mutated objects that can be in an invalid state and take the whole
+  application down (such as in languages like Ruby or Go).
+- Gleam offers syntax to make it easy to extract data out of custom types and
+  update data into new copies of custom types without ever mutating variables.
+  Go sometimes directly mutates references of simple values such as when using
+  pointers or global variables, though this is very much not idiomatic.
+- Gleam allows to rebind variables freely to make it easy to update data
+  structures by making a copy and binding it to the existing variable.
+- Go features a massive, powerful standard library centered around simple interfaces.
```
nii236, line 1384: These aren't really dot points, but paragraphs.

I would:

- Remove the dot points
- Move the last paragraph to the second paragraph so the Go and Gleam paragraphs are kept together

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1447 @@
...
+```
+
+Importing common types such as `gleam/order.{Lt, Eq, Gt}` or
+`gleam/option.{Some,None}` can be very helpful.
+
+## Architecture
+
+To iterate a few foundational differences (Go VS Gleam):
+
+1. Programming model: imperative VS functional immutable
```
nii236, line 1362: A common Go-ism is something like:

- Declare struct
- Declare constructor
- Declare methods on struct

It would be handy to see how you'd achieve this in Gleam. 

I know that for Gleam some people recommend a "context" struct that you pass in as the first argument which contains what would normally be Go struct fields: Database connections, loggers, HTTP clients etc.

cheatsheets/gleam-for-go-users.md
```diff
@@ -0,0 +1,1460 @@
...
+// a == 1
+// b == 2
+```
+
+#### Gleam
+
+In Gleam, `let` and `=` can be used for pattern matching, but you'll get
+compile errors if there's a type mismatch, and a runtime error if there's
+a value mismatch. For assertions, the equivalent `let assert` keyword is
+preferred.
```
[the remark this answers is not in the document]
GearsDatapacks: The compiler won't let you match a `let` pattern if it doesn't match all possible cases. `let assert` will cause a runtime error if the pattern doesn't match

PR #326, opened 2024-03-11
Fix small error in PHPs arrowFunction Syntax

thettler:
Arrow functions in PHP are written like: 
```php
$arrow = fn () =>  'Something';
```
the `fn` was missing in the docs.


inoas:
it is now fixed here https://github.com/gleam-lang/website/pull/327/files

lpil:
Thank you both!

PR #328, opened 2024-03-11
Use constructor property promotion

thekid:
Get rid of boilerplate - see https://wiki.php.net/rfc/constructor_promotion


lpil (approved):
Nice, thanks!

PR #329, opened 2024-03-11
Mac OS X -> macOS in the Installing page

fxn:
Easy one!

Apple renamed Mac OS X to macOS in June 2016, see [this post](https://512pixels.net/2016/06/history-of-macos/) for example.


lpil (approved):
Thank you

PR #333, opened 2024-03-13
Add install instructions for Winget

sitiom:
Gleam is available in Winget. See also:
- https://github.com/gleam-lang/gleam/pull/2699


lpil:
Excellent! Though I see the package is out-of-date. If we can get the package consistently up to date we can then add this to the website. Thank you

PR #334, opened 2024-03-13
Installation instructions for Alpine Linux

firefly-cpp:
**Gleam** has already been included in Alpine Linux repositories.


lpil (approved):
Thank you

PR #337, opened 2024-03-14
Fix some PHP code for the gleam-for-php-users guide

fitv:


lpil (approved):
Thank you

cheatsheets/gleam-for-php-users.md
```diff
@@ -291,7 +291,7 @@ pub fn distance(from x: Int, to y: Int) : Int {
   abs(x) - abs(y) |> abs()
 }
 // In some other function
-distance(from: 1, to: -2) // 3
+distance(from: 1, to: -2) // 1
```
inoas, line 294: but why should the distance from 1 to -2 be 1?

inoas: @fitv @lpil

inoas: nvm... just a bad name

PR #338, opened 2024-03-14
replace reference to old try keyword with explanation of use

Acepie:
this should close https://github.com/gleam-lang/gleam/issues/2690 I replaced the reference to the old `try` keyword with a short explanation of use based on the example that was already in the python cheatsheet.


lpil (approved):
Thank you

PR #339, opened 2024-03-14
Update gleam-for-php-users.md

inoas:
fix typo about return type, also rename function,

this works:

```gleam
import gleam/int
import gleam/io

pub fn absolute_distance(from x: Int, to y: Int) -> Int {
  int.absolute_value(x) - int.absolute_value(y) |> int.absolute_value()
}

pub fn main() {
// In some other function
  let dist = absolute_distance(from: 1, to: -2) // 1
  io.debug(dist)
}
```


lpil (approved):
Thank you

PR #342, opened 2024-03-16
Fix typo in Rust cheatsheet

catapillie:
The Rust/Gleam comparison cheatsheet had a missing space character in the operator comparison table, so this PR fixes it.


lpil (approved):
Thank you

PR #348, opened 2024-03-19
first pass at cheatsheet for scala

chadselph:
Hello,

For your consideration, I've written a cheatsheet for Scala programmers. I think by most measures Scala is the most mainstream fp language in industry and has a lot of similarity with Gleam. Both are highly influenced by ML family of languages but use a more C-style syntax.

I'm not sure how you decide which languages deserve a dedicated cheatsheet, and I realize this is more (basically untestable) code that needs to be maintained. So I will be understanding if you're not accepting these types of contributions.

There's also probably some mistakes, I based this off of the PHP cheatsheet but read several others to get ideas for what else to cover. I tried to keep in mind the audience should be Scala programmers looking into Gleam and not as much the other direction, so there's tons of omissions about Scala's huge type system.

Additionally, I think it would make sense to draw a comparison between the `use` expressions and Scala's `for` comprehensions, but I didn't see an obvious spot for that. Another minor thing to add is the `todo` vs Scala's `???` (which seems to have inspired `todo`) but that seems less important for real-world code.

### DIFF
--- a/cheatsheets/gleam-for-scala-users.md
+++ b/cheatsheets/gleam-for-scala-users.md
@@ -1,0 +1,1319 @@
+---
+layout: page
+title: Gleam for Scala users
+subtitle: Hello typesafe object-oriented functional programmers!
+---
+
+- [Comments](#comments)
+- [Variables](#variables)
+  - [Match operator](#match-operator)
+  - [Variables type annotations](#variables-type-annotations)
+- [Functions](#functions)
+  - [Exporting functions](#exporting-functions)
+  - [Function type annotations](#function-type-annotations)
+  - [Referencing functions](#referencing-functions)
+  - [Labelled arguments](#labelled-arguments)
+- [Operators](#operators)
+- [Constants](#constants)
+- [Blocks](#blocks)
+- [Data types](#data-types)
+  - [Strings](#strings)
+  - [Tuples](#tuples)
+  - [Lists](#lists)
+  - [Maps](#maps)
+  - [Numbers](#numbers)
+- [Flow control](#flow-control)
+  - [Case](#case)
+  - [Piping](#piping)
+  - [Try](#try)
+- [Type aliases](#type-aliases)
+- [Custom types](#custom-types)
+  - [Records](#records)
+  - [Algebraic Data Types](#algebraic-data-types)
+  - [Opaque custom types](#opaque-custom-types)
+- [Type Parameters](#type-parameters)
+- [Modules](#modules)
+  - [Imports](#imports)
+  - [Named imports](#named-imports)
+  - [Unqualified imports](#unqualified-imports)
+- [Architecture](#architecture)
+
+## Comments
+
+### Scala
+
+In Scala, comments are written with a `//` prefix.
+
+```scala
+// Hello, Joe!
+```
+
+Multi line comments may be written like so:
+
+```scala
+/*
+ * Hello, Joe!
+ */
+```
+
+In Scala, above `trait`, `object`, `class`, `def` declarations
+there can be `scaladoc` like so:
+
+```scala
+/**
+ * a very special trait.
+ */
+trait Foo {}
+
+/**
+ * A Bar class
+ */
+class Bar {
+
+  /**
+   *
+   * @param s describes the s param
+   */
+  def quux(s: String) = {}
+
+}
+
+```
+
+Documentation blocks (scaladoc) are extracted into generated API
+documentation and use Markdown format.
+
+### Gleam
+
+In Gleam, comments are written with a `//` prefix.
+
+```gleam
+// Hello, Joe!
+```
+
+Comments starting with `///` are used to document the following function,
+constant, or type definition. Comments starting with `////` are used to
+document the current module.
+
+```gleam
+//// This module is very important.
+
+/// The answer to life, the universe, and everything.
+const answer: Int = 42
+
+/// A main function
+fn main() {}
+
+/// A Dog type
+type Dog {
+  Dog(name: String, cuteness: Int)
+}
+```
+
+`//` comments are not used while generating documentation files, while
+`////` and `///` will appear in them.
+
+## Variables
+
+Scala has mutable and immutable variable bindings; in Gleam, all data is immutable but
+variable names can be reassigned in the same scope.
+
+### Scala
+
+```scala
+val size = 50
+val size = size + 100 // compile error
+val size = 1 // compile error
+
+var size2 = 50
+size2 = size2 + 100
+size2 = 1
+```
+
+Scala uses `val` for immutable bindings and `var` for mutable bindings.
+Lexical scoping can be done with braces or indentation, and values with the same name
+can shadow each other.
+
+```scala
+
+def foo() {
+  val size = 50
+
+  def bar() {
+    val size = 100
+  }
+}
+
+```
+
+
+### Gleam
+
+Gleam has the `let` keyword before its variable names.
+
+```gleam
+let size = 50
+let size = size + 100
+let size = 1
+```
+
+### Match operator
+
+Both Scala and Gleam can use the left hand side of an `=` as a pattern.
+
+#### Scala
+
+```scala
+val List(a, b) = List(1, 2)
+// a == 1
+// b == 2
+```
+
+#### Gleam
+
+In Gleam, `let` and `=` can be used for pattern matching, but you'll get
+compile errors if there's a type mismatch, and a runtime error if there's
+a value mismatch. For assertions, the equivalent `let assert` keyword is
+preferred.
+
+```gleam
+let #(a, _) = #(1, 2)
+// a = 1
+// `_` matches 2 and is discarded
+
+let assert [] = [1] // runtime error
+let assert [y] = "Hello" // compile error, type mismatch
+```
+
+Asserts should be used with caution.
+
+### Variables type annotations
+
+#### Scala
+
+Scala is a strongly typed language with pretty good type inference.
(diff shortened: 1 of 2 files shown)


cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+val size = 50
+val size = size + 100 // compile error
+val size = 1 // compile error
+
+var size2 = 50
+size2 = size2 + 100
+size2 = 1
+```
+
+Scala uses `val` for immutable bindings and `var` for mutable bindings.
```
lpil, line 133: Let's remove explanations and examples of `var` and the scoping rule below, as we don't need to teach Scala and there's no equivalent in Gleam.

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+// a == 1
+// b == 2
+```
+
+#### Gleam
+
+In Gleam, `let` and `=` can be used for pattern matching, but you'll get
+compile errors if there's a type mismatch, and a runtime error if there's
+a value mismatch. For assertions, the equivalent `let assert` keyword is
+preferred.
```
lpil, line 177: `let assert` is required, not preferred. `let` does not result in a runtime error as it does not permit value mismatches

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+```
+
+For methods with no arguments, you may omit the parenthesis
+
+```scala
+
+def noArgs = "Surprise."
+
+```
+
```
lpil, line 243:
```suggestion
```

We don't need to teach Scala

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+### Tuples
+
+Tuples are very useful in Gleam as they're the only collection data type that
+allows mixed types in the collection. They behave very similar to Scala's tuples,
+but are 0-index instead of 1 indexed when accessing by index.
+
+#### Scala
+
+```scala
+val myTuple = ("username", "password", 10) // Type is (String, String, Int) or Tuple3[String, String, Int]
```
lpil, line 562: Bit long this line!

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+```scala
+val list = List(2, 3, 4)
+val alsoList = 2 :: 3 :: 4 :: Nil // equivalent
+val list2 = 1 :: list // list of 1,2,3,4
+val 1 :: secondElement :: tail = list2 // secondElement = 2
+```
+
+#### Gleam
+
+Gleam has a `cons` operator that works for lists destructuring and
```
lpil, line 596: Gleam does not have a `cons` operator, it has a list prepend syntax

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+pattern matching. In Gleam lists are immutable so adding and removing elements
+from the start of a list is highly efficient.
+
+```gleam
+let list = [2, 3, 4]
+let list = [1, ..list]
+let [1, second_element, ..] = list
+[1.0, ..list] // compile error, type mismatch
```
lpil, line 604: Shouldn't this example match the Scala one above?

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+
+There is no map literal syntax in Gleam, and you cannot pattern match on a map.
+Maps are generally not used much in Gleam, custom types are more common.
+
+#### Scala
+
+```scala
+Map("key1" -> "value1", "key2" -> "value2")
+Map("key1" -> "1", "key2" -> 2) // compiles, but infers an unexpected type
+val myMap: Map[String, String] = Map("key1" -> "1", "key2" -> 2) // type error
```
lpil, line 625:
```suggestion
```

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+```
+
+#### Gleam
+
+```gleam
+1 / 2 // 0
+1.5 + 10 // Compile time error
+```
+
+You can use the gleam standard library's `int` and `float` modules to convert
```
lpil, line 664:
```suggestion
You can use the gleam standard library's `gleam/int` and `gleam/float` modules to convert
```

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+
+Case is one of the most used control flow in Gleam. It can be seen as a switch
+statement on steroids. It provides a terse way to match a value type to an
+expression. It is also used to replace `if`/`else` statements.
+
+Scala and Gleam both support guards, destructuring and disjoint union matching.
+
+#### Scala
+
+Scala has both `if`/`else` and `match` expressions.
```
lpil, line 680: Can we trim the Scala info in this section please

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+number match {
+  case 2 | 4 | 6 | 8 => "This is an even number"
+  case 1 | 3 | 5 | 7 => "This is an odd number"
+  case _ => "I'm not sure"
+}
+```
+
+#### Gleam
+
+The case operator is a top level construct in Gleam:
```
lpil, line 727: What's a top level construct?

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+```gleam
+case some_number {
+  0 -> "Zero"
+  1 -> "One"
+  2 -> "Two"
+  n -> "Some other number" // This matches anything
+}
+```
+
+As all expressions the case expression will return the matched value.
```
lpil, line 738: What does this mean?

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+
+if/elseif/else example:
+
+```gleam
+case True {
+  _ if is_admin == True -> "allow access"
+  _ if is_confirmed_by_mail == True -> "allow access"
+  _ -> "deny access"
+}
+```
```
lpil, line 768: Remove this please

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+
+Exhaustiveness checking at compile time, which is in the works, will make
+certain that you must check for all possible values. A lazy and common way is
+to check of expected values and have a catchall clause with a single underscore
+`_`:
```
lpil, line 773: Gleam has exhaustiveness checking

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+
+The callee block will be able to capture any exception raised in the block
+using a `try/except` set of blocks:
+
+```scala
+// callee block
+try {
+    println("this line will be executed and thus printed")
+    aFunctionThatFails()
+    println("this line will not be executed and thus not printed")
```
lpil, line 872: I think this should be 2 space

lpil (commented):
Thank you!

chadselph:
Thanks for the thorough review! I'll try to get these edits done soon

cheatsheets/gleam-for-scala-users.md
```diff
@@ -0,0 +1,1319 @@
...
+```
+
+In Scala, above `trait`, `object`, `class`, `def` declarations
+there can be `scaladoc` like so:
+
+```scala
+/**
+ * a very special trait.
+ */
+trait Foo {}
```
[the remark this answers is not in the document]
chadselph: yea I think these just came from using php as a starting point; will remove them

cheatsheets/gleam-for-scala-users.md
chadselph: I'll take a pass at re-writing this, but it's also included in the Elm, [Elixir](https://github.com/gleam-lang/website/blob/main/cheatsheets/gleam-for-elixir-users.md?plain=1#L106) and several other pages, so you might want to update those too.

cheatsheets/gleam-for-scala-users.md
chadselph: I've modified to Scala example to match this, as this is the example in the other cheatsheets. The last example (`[1.0, ..list] `) is a little problematic because the scala behavior is to upcast resulting list, which gets into OO type hierarchies none of which is related to the List data type

cheatsheets/gleam-for-scala-users.md
chadselph: This is another section that was re-used from the [PHP guide](https://gleam.run/cheatsheets/gleam-for-php-users/#case), and it still has this language on the site today.

I agree it doesn't really add anything here, especially for a scala audience since case will already be very intuitive to them.

cheatsheets/gleam-for-scala-users.md
chadselph: yeah this is from PHP as well and I think the author just meant the case expressions are expressions so

```gleam

let x = case y {
...
}
```

`x` is getting assigned to the "matched value". But I'm going to remove this because "matched value" really sounds more like the input (i.e. `y`) than the right hand side of the `->`

PR #349, opened 2024-03-21
chore: update installation instructions to mention homebrew for linux

isaacharrisholt:
Move Homebrew into a new 'cross-platform' section with `asdf`, and fix up some headers

### DIFF
--- a/getting-started/installing.md
+++ b/getting-started/installing.md
@@ -50,21 +50,31 @@ With [MacPorts](https://www.macports.org/) installed run the following:
 sudo port install gleam
 ```
 
-### Using the Nix package manager
+### Linux
+
+#### Using Homebrew
+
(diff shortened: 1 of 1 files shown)


getting-started/installing.md
```diff
@@ -31,9 +30,9 @@ The easiest way to install Gleam on Linux, Windows, and Apple macOS is to downlo
 prebuilt version of the compiler from the [GitHub release
 page](https://github.com/gleam-lang/gleam/releases).
 
-### macOS
+### UNIX Cross-platform
```
lpil, line 33: Brew isn't portable so this isn't quite right. I'm not sure asdf is either. Instead let's add brew to Linux.

isaacharrisholt: Done

PR #350, opened 2024-03-21
Minor improvements for "Gleam for Python users"

trag1c:
* the last Python code block was missing the language specifier
* made Python code blocks consistently use `py` (rather than mixing `py` and `python`)
* fixed syntax error in the lambda example
* updated type alias example to use 3.12 syntax (way more similar to Gleam now :))
* fixed incorrect assignment in the Person class example
* various style improvements (e.g. removing unnecessary `(object)` and `()` in class defs)


lpil (approved):
Thank you!

PR #351, opened 2024-03-22
feat: add link to installation instructions on homepage

isaacharrisholt:
Adds a link to installation instructions in the homepage hero. Previously it was difficult to find installation instructions.

Screenshot 2024-03-22 at 08 36 23

### DIFF
--- a/index.html
+++ b/index.html
@@ -11,7 +11,10 @@ preload:
   <b>Gleam</b> is a <b>friendly</b> language for building
   <b>type-safe</b> systems that <b>scale</b>!
 </div>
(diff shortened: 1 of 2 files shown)


lpil:
Hello! Please see previous discussions for why this is not currently possible.

isaacharrisholt:
> Hello! Please see previous discussions for why this is not currently possible.

Hey, I did look, but couldn't find anything. Could you point me in the right direction?

lpil:
Sorry I don't have the time to find them, but in short the install experience is too poor currently and we don't want to add friction to people trying Gleam and evaluating the language itself rather than the painful Erlang install experience.

PR #355, opened 2024-03-24
Add TypeScript cheatsheet

andyfleming:
Adds TypeScript cheatsheet (based primarily on the content of the PHP cheatsheet).

There's a ton of content that can be covered, but I think this is a good start. I've also included a list in an HTML comment at the end for some suggested topics for us to add next.

### DIFF
--- a/cheatsheets/gleam-for-typescript-users.md
+++ b/cheatsheets/gleam-for-typescript-users.md
@@ -8,27 +8,26 @@ subtitle: Hello TypeScripters!
 <!-- ----------------------------------------------------------------------- -->
 - [Comments](#comments)
 - [Variables](#variables)
-  <!-- - [Match operator](#match-operator) -->
-  <!-- - [Variables type annotations](#variables-type-annotations) -->
+- [Constants](#constants)
+- [Variables type annotations](#variables-type-annotations)
+- [Data types](#data-types) -->
+  - [Strings](#strings)
+  - [Numbers](#numbers)
+  - [Tuples, Arrays, Lists](#tuples-arrays-lists)
+  - [Objects and Records](#objects-and-records)
+  - [Maps](#maps)
+- [Destructuring Arrays and Tuples](#destructuring-arrays-and-tuples)
 <!-- - [Functions](#functions) -->
   <!-- - [Exporting functions](#exporting-functions) -->
   <!-- - [Function type annotations](#function-type-annotations) -->
   <!-- - [Referencing functions](#referencing-functions) -->
   <!-- - [Labelled arguments](#labelled-arguments) -->
 <!-- - [Operators](#operators) -->
-<!-- - [Constants](#constants) -->
 <!-- - [Blocks](#blocks) -->
-<!-- - [Data types](#data-types) -->
-  <!-- - [Strings](#strings) -->
-  <!-- - [Tuples](#tuples) -->
-  <!-- - [Lists](#lists) -->
-  <!-- - [Maps](#maps) -->
-  <!-- - [Numbers](#numbers) -->
 <!-- - [Flow control](#flow-control) -->
   <!-- - [Case](#case) -->
   <!-- - [Piping](#piping) -->
   <!-- - [Try](#try) -->
-<!-- - [Type aliases](#type-aliases) -->
 <!-- - [Custom types](#custom-types) -->
   <!-- - [Records](#records) -->
   <!-- - [Unions](#unions) -->
@@ -114,7 +113,6 @@ type Dog {
 
 `//` comments are not used while generating documentation files, while
 `////` and `///` will appear in them.
-
 ## Variables
 
 You can rebind variables in both languages.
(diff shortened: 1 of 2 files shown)


cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,794 @@
...
+if/else example:
+
+```gleam
+case is_admin {
+  True -> "allow access"
+  False -> "disallow access"
+}
+```
+
+if/elseif/else example:
```
samifouad, line 540: the first example is good, but i'm not sure if the second example is. perhaps a good way to showcase a more complex example of 'case' would be to do some nesting inside the True arm? eg. another case statement

```import gleam/io

pub fn main() {
  let x = 3
  let y = 2

  let result = case x {
    // Match specific values
    0 -> {
      case y {
        1 -> "hello"
        _ -> "good byee"
      }
    }
    1 -> "One"

    // Match any other value
    _ -> "Other"
  }
  io.debug(result)
}

**Review by samifouad — COMMENTED** [review]
awesome work! just a few comments i put inline

File: cheatsheets/gleam-for-typescript-users.md [hunk 2]
```diff
@@ -0,0 +1,794 @@
...
+```gleam
+let my_tuple = #("username", "password", 10)
+let #(_, pwd, _) = my_tuple
+io.print(pwd) // "password"
+// Direct index access
+io.print(my_tuple.0) // "username"
+```
+
+
+### Destructuring Arrays amd Tuples
```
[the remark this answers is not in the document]
andyfleming: Sorry, where?

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,794 @@
...
+const a: [number, number] = [1, 2]; // valid
+const b: [number, number] = [1, 2, 3]; // not valid
+const c = [1, 2]; // inferred type is number[], NOT a tuple
+const d: [string, number] = ["wow", 42];
+```
+
+#### Gleam
+
+Tuples are very useful in Gleam as they're the only collection data type that
+allows mixed types in the collection.
```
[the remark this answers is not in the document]
andyfleming: I'm not familiar with that syntax in Gleam. Can you give me an example?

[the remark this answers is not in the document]
samifouad: sure! i recently used it for a config builder with pipelines. but the underlying structure is a List of Tuples:

```import gleam/io

pub type KV =
  #(String, Int)

pub type Config =
  List(KV)

pub fn settings() -> Config {
  []
}

pub fn title(config: Config, title: Int) -> Config {
  [#("title", title), ..config]
}

pub fn description(config: Config, description: Int) -> Config {
  [#("description", description), ..config]
}

pub fn main() {
  let output = settings()
  |> title(1)
  |> description(2)
  
  io.debug(output)
}
```

If run the above should output:

`[#("description", 2), #("title", 1)]`

so technically the tuple is doing the multi-type work here, but using tuples like this gives people coming from TypeScript who are used to objects some more flexibility

cheatsheets/gleam-for-typescript-users.md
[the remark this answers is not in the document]
samifouad: Arrayss "amd" Tuples

cheatsheets/gleam-for-typescript-users.md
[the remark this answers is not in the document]
andyfleming: Wouldn't a record be better here?

```gleam
import gleam/io
import gleam/list

pub type IntOrFloat {
  AnInt(Int)
  AFloat(Float)
}

pub type NumList =
  List(IntOrFloat)

pub fn nums_list() -> NumList {
  []
}

pub fn main() {
  let output = nums_list()
  |> list.append([AnInt(1)])
  |> list.append([AFloat(2.5)])
  
  io.debug(output) // [AnInt(1), AFloat(2.5)]
}
```

Although, I'm not sure if this is actually the recommended way to do something like this.

cheatsheets/gleam-for-typescript-users.md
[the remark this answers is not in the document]
andyfleming:
```suggestion
### Destructuring Arrays and Tuples
```

Apparently I'm blind, haha. Thanks!

cheatsheets/gleam-for-typescript-users.md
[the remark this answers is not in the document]
andyfleming: I think these examples may be too confusing for newcomers.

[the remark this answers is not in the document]
samifouad: agreed. might be worth just a mention to research into. in the future maybe even a link to a gist or example page ([like this one from elm](https://elm-lang.org/examples)) once a good one exists

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,826 @@
...
+const arr6: number[] | string[] = [1, 2, 3]; // Valid. Array of numbers OR array of strings
+const arr7: number[] | string[] = ["one", "two", "three"]; // Valid
+const arr6: number[] | string[] = ["one", 2, "three"]; // Invalid
+```
+
+#### Gleam
+
+Gleam has a `cons` operator that works for lists destructuring and
+pattern matching. In Gleam lists are immutable so adding and removing elements
+from the start of a list is highly efficient.
```
coder0107git, line 261: It might be good to mention that lists are singly-linked so they're only efficient when adding or removing from the start of a list and that elsewhere is less efficient (has to traverse all previous nodes to get to the right one).

cheatsheets/gleam-for-typescript-users.md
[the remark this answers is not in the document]
andyfleming: Yeah, that could be cool

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,826 @@
...
+  - Compiling Gleam to JavaScript
+  - Exporting TypeScript types from Gleam
+  - Labelled arguments (vs options objects in TypeScript)
+  - Classes in TypeScript
+  - Import aliases
+  - Unions
+  - Operators
+  - Promises
+  - Concurrency patterns 
+  - Examples of array/list operations
```
bondiano, line 824: also we probably should add generics section

andyfleming, line 824:
```suggestion
  - Examples of array/list operations
  - Generics
```

cheatsheets/gleam-for-typescript-users.md
coder0107git: Has this been added in a local commit and just not been pushed to Github yet?

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+scenarios, using a case for instance:
+
+```gleam
+case parse_int("123") {
+  Ok(i) -> io.println("We parsed the Int")
+  Error(e) -> io.println("That wasn't an Int")
+}
+```
+
+In order to simplify this construct, we can use the `try` keyword that will:
```
lpil, line 758: This feature hasn't existed in some time!

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+var promisedString;
+ 
+// You can specify an HTML Element with DOM properties
+/** @type {HTMLElement} */
+var myElement = document.querySelector(selector);
+element.dataset.myData = "";
+```
+
+Documentation blocks (docblocks) are extracted into generated API
+documentation.
```
lpil, line 66: We don't need to teach TypeScript features here

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+
+### Gleam
+
+Gleam also has the `let` keyword before its variable names. However, it needs to be used explicitly each time when rebinding/reassinging a new value.
+
+```gleam
+let size = 50
+let size = size + 100
+let size = 1
+```
```
lpil, line 117: This isn't quite right. In TypeScript you can mutate bindings, while in Gleam you can shadow them. This is a big difference so we should be sure to explain the difference here.

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+  the_answer
+}
+```
+
+They can also be marked public via the `pub` keyword and will then be
+automatically exported.
+
+```gleam
+pub const the answer = 42
+```
```
lpil, line 147: Gleam constants are largely unrelated to TypeScript `const`. The comparison should be between in-function assignments in each language, and module level assignments in each language.

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+let some_list: List(Int) = [1, 2, 3]
+let some_string: String = "Foo"
+let another_string = "Bar" // valid, inferred without explicit type annotation
+```
+
+Gleam will check the type annotation to ensure that it matches the type of the
+assigned value. It does not need annotations to type check your code, but you
+may find it useful to annotate variables to hint to the compiler that you want
+a specific type to be inferred.
+## Data types
```
lpil, line 175:
```suggestion

## Data types
```

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+
+/// A Dog type
+type Dog {
+  Dog(name: String, cuteness: Int)
+}
+```
+
+`//` comments are not used while generating documentation files, while
+`////` and `///` will appear in them.
+## Variables
```
lpil, line 97:
```suggestion

## Variables
```

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+### Strings
+
+In TypeScript (JavaScript), strings are represented fundamentally as sequences
+of UTF-16 code units. They can be defined with single quotes, double quotes,
+or back-ticks (also known as template strings). Interpolation is supported in
+template literal strings.
+
+In Gleam all strings are UTF-8 encoded binaries. Gleam strings do not allow
+interpolation, yet. Gleam however offers a `string_builder` via its standard
+library for performant string building.
```
lpil, line 186: This representation info is only true on the Erlang platform, and let's remove the word "yet" as it implies it is planned.

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+
+const arr5: Array<string|number> = [1, "2", 3]; // Valid. Array of numbers and/or strings
+const arr6: number[] | string[] = [1, 2, 3]; // Valid. Array of numbers OR array of strings
+const arr7: number[] | string[] = ["one", "two", "three"]; // Valid
+const arr6: number[] | string[] = ["one", 2, "three"]; // Invalid
+```
+
+#### Gleam
+
+Gleam has a `cons` operator that works for lists destructuring and
```
lpil, line 259: Gleam does not have a cons operator, and prefer the term "pattern matching" over "destructuring", especially in the context of JS as the semantics are different.

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+const a: [number, number] = [1, 2]; // valid
+const b: [number, number] = [1, 2, 3]; // not valid
+const c = [1, 2]; // inferred type is number[], NOT a tuple
+const d: [string, number] = ["wow", 42];
+```
+
+#### Gleam
+
+Tuples are very useful in Gleam as they're the only collection data type that
+allows mixed types in the collection.
```
lpil, line 286: Custom types also support containing values of mixed types

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+```gleam
+let my_tuple = #("username", "password", 10)
+let #(_, pwd, _) = my_tuple
+io.print(pwd) // "password"
+// Direct index access
+io.print(my_tuple.0) // "username"
+```
+
+
+### Destructuring Arrays and Tuples
```
lpil, line 297: We don't use the term destructuring or have arrays. What should we call this?

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+For lists, the `list.split` method may be used.
+
+```gleam
+import gleam/io
+import gleam/list
+
+pub fn main() {
+  let ints = [11, 22, 33]
+
+  let #(a, rest) = list.split(ints, 1)
```
lpil, line 343: This is invalid code and will be rejected by the compiler.

We do not recommend using list.split like this either. It's not a useful pattern in Gleam so best not to confuse them by showing it.

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+}
+
+obj.a; // 1
+obj['a']; // 1
+obj.c.neat; // true;
+```
+
+There are several ways to types objects in TypeScript. They provide a Record type,
+however, it is not a true record. Though, there is [a proposal to add true records
+and tuples](https://github.com/tc39/proposal-record-tuple) to JavaScript.
```
lpil, line 373: We don't need to teach TS in this guide, we can focus on the Gleam

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+}
+```
+
+### Sets
+
+A set is a collection of unique members. Both TypeScript and Gleam support sets.
+
+#### TypeScript
+
+A set in TypeScript can a set can have multiple types.
```
lpil, line 421: Not quite- it has a single value type but TypeScript has subtyping. This would be true of everything in this guide so let's not bring it up specifically for sets.

The main difference is that in Gleam these data structures are immutable

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+
+set.new()
+|> set.insert(2)
+|> set.contains(2)
+// -> True
+```
+## Functions
+
+Both TypeScript and Gleam have first-class functions. Both support functions as values,
+as well as higher-order functions, partial application, currying, and similar patterns.
```
lpil, line 450: Neither Gleam or TypeScript has currying built in, you have to implement it yourself.

cheatsheets/gleam-for-typescript-users.md
```diff
@@ -0,0 +1,827 @@
...
+}
+```
+
+### Switch and Case
+
+#### TypeScript
+
+TypeScript has simple switch statements that allow for some basic matching of a value.
+
+There is a language proposal for [a pattern matching syntax](https://github.com/tc39/proposal-pattern-matching).
```
lpil, line 588: Let's not talk about future potential additions to TS

cheatsheets/gleam-for-typescript-users.md
andyfleming: I was just translating to the equivalent from this cheatsheet: https://github.com/gleam-lang/website/blob/main/cheatsheets/gleam-for-php-users.md?plain=1#L58-L82

I can drop that section though.

cheatsheets/gleam-for-typescript-users.md
andyfleming: Something like this?

> Gleam also has the `let` keyword before its variable names. However, variables cannot be reassigned. The names can be used again by binding with another `let` statement, but the values themselves are not changed or mutated in any way. The values they reference are immutable.

cheatsheets/gleam-for-typescript-users.md
andyfleming:
```suggestion
In Gleam strings are UTF-8 encoded binaries (on the Erlang platform). Gleam strings do not allow
interpolation. Gleam however offers a `string_builder` via its standard
library for performant string building.
```

PR #361, opened 2024-04-04
Add installation link to page header

davisuga:


inoas:
I think semantically it should be in the 2nd nav block

Issue #368, opened 2024-04-13
Recommend Semver Compatible Versioning over ~> in Docs


Pi-Cla:
Since we are trying to encourage people to [move away from shorthands](https://github.com/gleam-lang/gleam/issues/2976#issuecomment-2053589370) such as `~>`, this then brings up the question of what should be "recommended" as an alternative instead. Right now the [docs about gleam.toml](https://gleam.run/writing-gleam/gleam-toml/) still use `~>` in all of their examples so that should be changed to something else.

I think we could encourage the usage of [semver compatible versioning](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) like so:
# Old
```toml
# The Hex packages the project needs to compile and run (optional)
# Uses the Hex version requirement format
# https://hexdocs.pm/elixir/Version.html#module-requirements
[dependencies]
gleam_stdlib = "~> 0.18"
gleam_erlang = "~> 0.2"
gleam_http = "~> 2.1"
# Local dependencies can be specified with a path
my_other_project = { path = "../my_other_project" }

# The Hex packages the project needs for the tests (optional)
# These will not be included if the package is published to Hex.
# This table cannot include any packages that are already found in the
# `dependencies` table.
[dev-dependencies]
gleeunit = "~> 0.3"
gleam_bitwise = "~> 0.3"
```
# New
```toml
# The Hex packages the project needs to compile and run (optional)
# Uses the Hex version requirement format
# https://hexdocs.pm/elixir/Version.html#module-requirements
[dependencies]
gleam_stdlib = ">= 0.18 and < 0.19"
gleam_erlang = ">= 0.2 and < 0.3"
gleam_http = ">= 2.1 and < 3"
# Local dependencies can be specified with a path
my_other_project = { path = "../my_other_project" }

# The Hex packages the project needs for the tests (optional)
# These will not be included if the package is published to Hex.
# This table cannot include any packages that are already found in the
# `dependencies` table.
[dev-dependencies]
gleeunit = ">= 0.3 and < 0.4"
gleam_bitwise = ">= 0.3 and < 0.4"
```

PR #369, opened 2024-04-13
Replace ~> with Semver Compatible Versioning

Pi-Cla:
See: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio
Resolves: #368


lpil (commented):
Hello! Thank you for this. You'll need to add `.0` to the end of the versions for this to be valid.

PR #372, opened 2024-04-16
Fix Elm cheatsheet's Operators table and other bits and bobs

ianmjones:
I noticed that the Elm cheatsheet's Operators table was not formatted correctly, and failed to show the `||` and `|>` operators correctly.

https://gleam.run/cheatsheets/gleam-for-elm-users/#operators

I've fixed it up:

2024-04-16-21-43-10_1876x1846_scrot

While there, I also noticed that some things were out of date, in particular some links to the tour, references to the JavaScript target were missing in some places, and the old `external` keyword was being used.


lpil (approved):
Thank you

PR #374, opened 2024-04-21
Rotate all the Lucy's

graphiteisaac:
This is a pretty straight-forward PR, I just downloaded the existing Lucy images and rotated them in Figma, and removed the paragraph specifying you need to manually rotate. This also drops the need to rotate the logo in CSS :)


writing-gleam/command-line-reference.md
MAHcodes, line 1: did you delete this file by accident?

graphiteisaac: Yes I did, woopsies. I even noticed it before but forgot to restore it before pushing 😭

graphiteisaac:
I accidentally deleted that file by running `make build` I think. Fixed now so this PR should be theoretically good to go.

lpil (commented):
Thank you!! The Lucy on the hero of the homepage is now rotated too much to the left and not rotated right enough on hover

graphiteisaac:
Good catch! I should have compared my local copy to prod, have now fixed this and it matches up pretty well. The sizing is slightly different but it's not very noticeable and is a result of the image size changing to accommodate for the rotation

lpil (approved):
ooooh and a new flag Lucy !! Awesome!! Thank you

PR #362, opened 2024-04-04
Invert Fly.io's logo on all news posts + Added news page styling #354

Dhuds1:
## Non merge related
Hello, I am really new to this community and I'm very much in love with y'alls! I'm currently in a college and our most recent assignment was to contribute to an open source community. I've been looking for a way to contribute to this community for a while, and I noticed a few things that I thought I could help with!

### Fly.io logo to inverted version
issue #354 
I added the fly.io's inverted logo to the `images/sponsors folder`, and named it `fly-inverted.svg`. I went through each news post and updated the `fly.svg` to `fly-invert.svg`

### News.html styling
Afterwards, I thought that the news.html page was a little hard on the eyes, and that there was a lack of typographic hierarchy with the text which makes it hard to figure out which text is most to least important. I changed the `publish date, and author` text to`var(--color-code-grey)` and text size to `var(--font-size-small)`. Afterwards, I added a hover effect on the `{{ post.title }}` to help clarify what is a link. I then changed the `li { display }` to better space its content. I then added a `li::before` to create a bar of `var(--color-faff-pink)` to the left size of the articles. This helps "block" the articles, and add a little visual interest.  Oh, I forgot, I also added the same lucy-hero effect to the Lucy in the footer, for some ✨ Spice ✨

### News.html alterations
Inside the each post `<li>` I added custom data fields, `post-tags` and `post-author` with the idea that maybe later on you'd want to implement some sort of user-filtering. I also added an `<a href="{{ post.author-link }}">` tag to `{{ post.author }}` because I figured maybe someone would want to check out more about the author?

Sorry for the wall of text, this is my first time creating a PR and I wanted to be concise.

## TL:DR;
- made `fly.io` inverted on news articles
- added typographical hierarchy to posts in `news.html`
- added visual interest to posts in `news.html` 
- added `hover` on `{{ post-title }}`
- added `hover rotation and svg change` on `lucy` footer version
- added author link to `{{ post.author }}`
- added `post-tags ` and `post-author` data-attributes `into posts

PR #376, opened 2024-04-22
fix: :bug: fixes fly io logo for post 2002-11-24-v0.25-xxx

### LINKED ISSUE #354: Fly.io logo looks bad on dark background
image

https://gleam.run/news/v0.25-introducing-use-expressions/

I'm not sure if this is that important or not. I randomly clicked on a link and saw this. It doesn't look good 🤷‍♂️

xxKeefer:
corrected src atribute for image tag

#354

pretty sure this can close #354, all the hard work was completed in #362

the only reference to the light theme logo is on the home page in the top sponsor banner which looks good, the other references are all in posts which are on a dark back ground, it was just this instance that had been missed.


lpil (approved):
Thank you!

PR #377, opened 2024-04-23
Add temporary openSUSE installation instructions

Pi-Cla:
These will be replaced with just
```
# zypper install gleam
```
once Gleam is accepted into the official repository. I don't know why it is taking so long though, openSUSE is usually good with accepting new packages quickly....


lpil (approved):
Lovely! Thank you

PR #382, opened 2024-05-03
Update openSUSE installation instructions

Pi-Cla:
It finally got in after so long thanks to Matwey Kornilov of devel:languages:erlang. BEAM language solidarity!


lpil (approved):
Brilliant! thank you again 💜

PR #384, opened 2024-05-10
Fix Wasm logo and add JS, Erlang, Godot and Nix

DanielleMaywood:
The Wasm logo is off-centre, also adds a JS Lucy.


DanielleMaywood:
Added Nix and Erlang

branding.html
```diff
@@ -84,6 +84,11 @@ <h4>✨ How clumsy as Lucy?</h4>
   <li><img src="/images/lucy/lucygay.svg" alt="Lucy gay" /></li>
   <li><img src="/images/lucy/lucybi.svg" alt="Lucy bi" /></li>
   <li><img src="/images/lucy/lucywasm.svg" alt="Lucy Wasm" /></li>
+  <li><img src="/images/lucy/lucyjs.svg" alt="Lucy JS" /></li>
+  <li><img src="/images/lucy/lucyerl.svg" alt="Lucy Erlang" /></li>
+  <li><img src="/images/lucy/lucynix.svg" alt="Lucy Nix" /></li>
+  <li><img src="/images/lucy/lucyflake.svg" alt="Lucy Flake" /></li>
+  <li><img src="/images/lucy/lucygodot.svg" alt="Lucy Godot" /></li>
```
lpil, line 91: It would be nice to credit you as the artist somehow. Alt text perhaps?

lpil (commented):
These are the cutest things ever

branding.html
DanielleMaywood: I personally don't mind if I'm credited or not.

We should probably ask @isaacharrisholt if they want crediting for `lucyflake.svg` as all I did was convert their design into an svg.

isaacharrisholt: I also don't mind, but if I do, it's only fair that @DanielleMaywood gets joint credit

lpil: I want to record it! 😁

isaacharrisholt:
Side note, I've just noticed a typo in the FAQ, where it says 'How clumsy **as** Lucy?'. Is it worth just adding that small fix to this PR, or should I create a new one?

lpil:
Either or works for me!

lpil (approved):
💜 💜 💜

PR #385, opened 2024-05-11
Rename mentions of the old Map type to Dict

PastMoments:
As in title, many language tours still hadn't been updated to refer to the type as Dict.

This PR just renames and slightly rephrased those sections.


lpil (approved):
Thank you

PR #391, opened 2024-05-16
add lucy elixir logo

okkdev:
Modified version of the Elixir Logo found in the nerd font.
https://github.com/ryanoasis/nerd-fonts/blob/master/src/svgs/elixir_nf.svg


lpil (approved):
Thank you!!

PR #392, opened 2024-05-26
Update gleam-for-elixir-users.md

MiniAppleTheApple:
# Todo

- [ ] Flow Control 
  - [ ] Case
  - [ ] Try
- [ ] Type aliases
- [ ] Custom types
  - [ ] Records
  - [ ] Unions
  - [ ] Opaque custom types
- [ ] Modules
  - [ ] Imports
  - [ ] Nested modules
  - [ ] First class modules

### DIFF
--- a/cheatsheets/gleam-for-elixir-users.md
+++ b/cheatsheets/gleam-for-elixir-users.md
@@ -661,29 +661,29 @@ pub fn main() {
 
 ## Patterns
 
-Same as Elixir, Gleam has pattern matching. Which is used for matching complex structured data.
+Same as Elixir, Gleam has pattern matching, which is used for matching complex structured data. In Gleam we use `as` to name the variable, same as using `=` in Elixir.
 
 #### Gleam
 
 ```gleam
 let list = [1, 2, 3]
 
-let assert [first, second .. rest] = list
-let assert [1, second .. rest] = list
+let assert [first as 1, second .. rest] = list
 
-let assert Foo(foo: 10, bar: bar) = Foo(foo: 10, bar: 20)
-let assert 20 = bar
+let assert Foo(foo: 10 as foo, bar: 20 as bar) = Foo(foo: 10, bar: 20)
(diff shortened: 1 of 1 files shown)


cheatsheets/gleam-for-elixir-users.md
```diff
@@ -659,3 +659,32 @@ pub fn main() {
...
+## Patterns
+
+Same as Elixir, Gleam has pattern matching, which is used for matching complex structured data. In Gleam we use `as` to name the variable, same as using `=` in Elixir.
+
+#### Gleam
+
+```gleam
+let list = [1, 2, 3]
+
+let assert [first as 1, second .. rest] = list
```
lpil, line 671:
```suggestion
let assert [1 as first, second, ..rest] = list
```

cheatsheets/gleam-for-elixir-users.md
```diff
@@ -659,3 +659,32 @@ pub fn main() {
...
+
+let assert Foo(foo: 10 as foo, bar: 20 as bar) = Foo(foo: 10, bar: 20)
```
lpil, line 673: No "foo" please, it's a rather depressing WW2 reference.

lpil:
Would be good to explain that Gleam has exhaustiveness checking here

cheatsheets/gleam-for-elixir-users.md
inoas: I did not know till now, no more foo/foobar anymore for me then.

MiniAppleTheApple: Sure, thanks for suggesting

MiniAppleTheApple:
The idea of using let assert is to let the readers know the equivalent feature in Gleam. However, I will address the exhaustiveness checking you mentioned and explain to the readers a more "Gleam" way of thinking instead of the "Elixir" way.

PR #394, opened 2024-05-26
Fix typos on `gleam-for-elm-users.md`

korokd:


korokd:
Hijacking my own PR to say that it's really nice to have these comparisons between languages that might be familiar to those interested in the language.

lpil (approved):
Ah! Thank you!!

PR #397, opened 2024-05-29
Add clarification about using Mason in Neovim

justinrassier:
This is an attempt to add some clarification for those that are getting started with Gleam in Neovim. 

Mason recommended almost everywhere now days to get started with Neovim and its LSP integration. Because of this, new users see the "deprecated" message in Mason about Gleam which adds friction and confusion to the getting started process.

I have also added a PR to Mason to direct people to this page to get more information https://github.com/mason-org/mason-registry/pull/5904

### DIFF
--- a/language-server.md
+++ b/language-server.md
@@ -60,6 +60,13 @@ If you are using [`nvim-treesitter`](https://github.com/nvim-treesitter/nvim-tre
 you can run `:TSInstall gleam` to get syntax highlighting and other tree-sitter
 features.
 
+### What About Mason?
+Gleam's toolchain comes bundled with everything in a single binary, including the LSP. 
+This means if you install Gleam using Mason, Neovim gets a second version of Gleam, which 
+can lead to version conflicts between your editor and your actual build.
+
+Because of this, it is recommended to just add the manual configuration line above using `lspconfig`.
+
 ## VS Code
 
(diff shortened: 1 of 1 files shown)


lpil:
That mason package is the bane of my existence atm. Is there a way to remove it?

justinrassier:
> Is there a way to remove it?

🤔 I'm not sure. I can look a little deeper and see if I can change the PR in the mason-registry to remove it!

justinrassier:
I didn't see any information about removing a package, but I tried it locally and it seemed to work just fine. So I created a new PR to remove it from the registry completely 

https://github.com/mason-org/mason-registry/pull/5911

It probably wouldn't hurt for you to pop in there and give your endorsement, as some random guy deleting things from a registry may not be well received :)

If there is any follow-up work, I'd be happy to work with them after they know this is a legitimate request.

justinrassier:
Hey hey! It looks like they merged it in this morning, and I just checked it out. I no longer see gleam in the list of LSPs for Mason!

I am wondering if a clarification like this still may be nice, considering many people use Mason and may no understand why it is not in the list?

lpil:
Awesome! I don't think we need to make a comment about Mason, we don't for the others and we say how to install the LS just above.

justinrassier:
Sounds good! Closing the PR!

🥳

PR #398, opened 2024-06-03
Avoid flickering in logo hover animation

lydell:
As I was about to try Gleam, Lucy shook their head as if I was about to make a big mistake.

https://github.com/gleam-lang/website/assets/2142817/21d7609d-2f27-4169-be64-98c986e41373

This PR adds a container around Lucy, with `overflow: hidden` so the rotation happens without changing the hover boundaries. Also, the hover effect no longer starts in random distances from the logo, by changing some padding to margin.

(For a language that took inspiration from Elm, it’s funny you even copied [having a somewhat glitchy hover effect on your website](https://twitter.com/SimonLydell/status/1756691086892797991) 😄 )

### DIFF
--- a/_includes/page-header.html
+++ b/_includes/page-header.html
@@ -26,12 +26,13 @@
   <div class="hero">
     <div class="content">
       {% if include.image %}
-      <img
-        data-show-pride
-        class="hero-lucy wide-only"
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you!!!

PR #406, opened 2024-07-11
Add a link for Gleam Weekly

CrowdHailer:


lpil (approved):
Thank you

lpil (commented):
Oh, sorry. Could you put this on the community page instead

CrowdHailer:
I think you mean alongside things here 
image
If so do you have a place you source the svg's from. I would suggest getting an email icon SVG?

Separately to that I think having it next to discord, whereever that might be i.e. here in the footer, makes sense because I think it always catches the people for whom discord is too high frequency. (I spotted a Gleam weekly link in the Gloogle footer which is what gave me the idea)

lpil:
I'd probably add a section below them as it's not as important as either of those two forums

CrowdHailer:
I'm not entirely sure how you would like this to be added here.

CrowdHailer:
How does this look?

lpil (approved):
Sorry for the slow reply, thank you!

PR #407, opened 2024-07-12
Clarify / simplify OpenBSD instructions

VlkrS:
gleam has been available in OpenBSD ports for two releases now, so there's no point in highlighting that it's available in -current.


lpil (approved):
Oh wonderful! Thank you.

PR #408, opened 2024-07-14
Fix some Python code segments and simplify grammar

broad-well:
For the Gleam for Python users cheatsheet, I made sure that all the Python code segments run as intended and simplified some grammar. I also removed a few extra spaces and clarified what we mean by "the `return` keyword is mandatory" for functions defined with `def`.


lpil (approved):
Thank you!

PR #419, opened 2024-07-26
Adds heading for Elixir Cheatsheet ToC to link to

justynhunter:
Right now, on the live site the 'Records' item in the elixir cheat sheet table of contents does not work because there is no `### Records` in the document.


lpil (approved):
Thank you

PR #420, opened 2024-07-29
remove-foo-bar-quux-batz-kek-references

inoas:
```
Foo => Wibble
Bar => Wabble
Batz => Wobble
Quux => Wubble
Kek => Webble
```

now I know kek is not necessarily a foo/bar thing but, nevermind.

### DIFF
--- a/cheatsheets/gleam-for-elixir-users.md
+++ b/cheatsheets/gleam-for-elixir-users.md
@@ -428,7 +428,6 @@ fn main() {
 }
 ```
 
-
 ## Blocks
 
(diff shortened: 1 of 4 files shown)


inoas:
While doing this I noticed that the memory effect of foo/bar/quux is much more separated than wibble, wabble, wobble. For now this replacement is OK but I will think of different placeholders that are more distributed among the alphabet and more distinct to remember and differentiate.

lpil (approved):
Thank you!!!

PR #421, opened 2024-08-02
Proofread the Python tutorial

sobolevn:
Basically, I only change one major example about `NewType`. Other changes are more like gotchas.

For example:
- any string can be a docstring (except for `f` strings), not only `"""` ones
- type annotations also exist on python2 via type comments
- It is common to add `typing.Final` to module / class level constants
- etc :)


lpil (approved):
Thank you

PR #424, opened 2024-08-02
Fix Typo in "Supercharged Labels"

zsarge:
Currently, [the post](https://gleam.run/news/supercharged-labels/) reads:

> The shorthand can be used for labelled record fields and in pattern matching too.
> ```gleam
> pub fn get_year(data: Date) -> Year {
>   let Date(year:, ..) = date
>   year
> }
> ```

I haven't ran this code, but I would anticipate that the parameter was supposed to be "date" and not "data".

Feel free to disregard this PR if I'm missing something.


lpil (approved):
Thank you

PR #426, opened 2024-08-03
Update 2024-08-02-supercharged-labels.md

Tuna4242:
Changed a tiny tiny typo under the case correction section from you're to you'll, feel free to shut me down and explain how the previous instance was more correct! This is my first ever github change so I could be wrong.


lpil (approved):
Thank you very much!!!

PR #431, opened 2024-08-08
Added rotation for lucy in a header

lixelv:
I added a rotation for the Lucy logo in the header. It now behaves the same way as the logo in the hero and footer sections. I've added class footer-lucy to header lucy logo img.

### DIFF
--- a/_includes/page-header.html
+++ b/_includes/page-header.html
@@ -6,6 +6,7 @@
           <img
             src="/images/lucy/lucy.svg"
             alt="Lucy the star, Gleam's mascot"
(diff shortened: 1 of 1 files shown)


_includes/page-header.html
```diff
@@ -6,6 +6,7 @@
           <img
             src="/images/lucy/lucy.svg"
             alt="Lucy the star, Gleam's mascot"
+            class="footer-lucy"
```
lpil, line 9: This says it's the footer, but it's the header!

PR #444, opened 2024-10-03
fix: Updated the image version to 1.5.1

AsymetricData:
The deployment image version wasn't sync to current latest version available !


lpil (approved):
Thank you

PR #447, opened 2024-10-29
Add a bunch of Lucys

okkdev:
Adds the Lucys which are already available as Discord emojis, as well as a "tiny" lucy optimized for small use cases.


lpil (commented):
Could you remove those other than the tiny ones please. 🙏 

What do they look like when rendered on the web page? Are they suitably small?

okkdev:
Oh sure, should I add the other ones in a separate PR or not at all?

Right now all of them have the same size, should I display the tiny ones small?

lpil:
Not added, they don't match the Gleam branding or tone so they're not suited to be added to the official assets.

Yes please, the tiny ones should be shown at the size at which they should be used.

okkdev:
Oh my bad! I was asked to PR them after they got added on the Discord, I just didn't get around to doing it till now.

Sure, they'll be very tiny and easy to miss as 16x16px tho!

okkdev:
like this?
image

lpil:
ahah gosh that looks very silly. Yes perfect!

okkdev:
This should be it :)

lpil (approved):
Thank you!

PR #449, opened 2024-10-31
Changed the wording into the past tense.

pastc:
This is because fly is no longer sponsoring the gleam project.


lpil (approved):
Thank you!

PR #450, opened 2024-11-03
make gleam homepage title a bit more specific

george-grec:
I think adding a title that specifies "Language" looks better in search engines.

How it looks on the web now:
It just says "Gleam"
image

**Comparisons**
Competing search result for "gleam":
image

golang:
image

Rust:
image

Haskell:
image

I chose to go with "Gleam Language" (just like Haskell does) because the name still fits nicely in the title bar. Other pages remain unchanged, they still append "- Gleam" after their title.

### DIFF
--- a/_layouts/layout.html
+++ b/_layouts/layout.html
@@ -2,7 +2,9 @@
 <html lang="en-GB">
   <head>
     <!-- prettier-ignore -->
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you!

PR #452, opened 2024-11-05
Update gleam-toml.md to include the new repository.path option

richard-viney:


inoas:
I think it would be nice if the docs about gleam.toml where somehow part of the compiler and listable via `gleam help gleam.toml` or `gleam readme gleam.toml` or something

lpil (approved):
Thank you! Will hold off on merging until it is released

PR #455, opened 2024-11-18
Fix missing words in Gleam 1.6 announcement

philpax:
Noticed these while reading through the announcement, so made a few quick fixes ^_^

I also noticed a few other things that would benefit from a grammar pass (mostly around using commas to more clearly delineate sentence clauses), but I figured I should keep this PR relatively unobjectionable and address the most obvious issues.


lpil (approved):
Oops, thank you!

PR #459, opened 2024-11-27
add social lucys

aslilac:
Add two variants of Lucy, fitted for use in "social" contexts (like the reddit and the discord server)

- One with a pink background
- One with a pride background


lpil (approved):
Heck yeah!!!

PR #461, opened 2024-12-04
Fix Zerops Guide

nermalcat69:
Changes Made
- Removed subdomain access from import yml
- Fixed Typo for service name (it's supposed to be Gleam Service)
- Added the command for for enabling subdomain with zcli


lpil (approved):
Thanks!

PR #467, opened 2025-01-05
Fix typo in v1.7.0 announcement

lue-bird:
Fixes a small typo.

IMO it would also read nicer as
```diff
- Triggering the code action result in the function-capture being expanded to the full anonymous function syntax:
+ Triggering the code action expands the function-capture to the full anonymous function syntax
```
but it's not a big difference.


lpil (approved):
Thank you

PR #471, opened 2025-02-04
Update roadmap

giacomocavalieri:
- I've moved features that shipped with 1.7 in their own section
- Moved features that will be shipping with 1.8 in the "in progress" section
- And also added a couple of notable additions that I could think of being worked on


lpil (commented):
Thank you!

lpil (approved):
Oops, wrong button

PR #473, opened 2025-02-07
Update roadmap for new release

giacomocavalieri:
I've updated the roadmap to include v1.8, I've also changed the double quotes in the title of the v1.8 announcement post as the opening one looked odd:
image


lpil (approved):
Thank you!

PR #477, opened 2025-02-16
Add example of git dependencies to gleam toml documentation

GearsDatapacks:
This PR goes alongside https://github.com/gleam-lang/gleam/pull/4247, and documents the now valid use of git dependencies. This should probably only be merged once the next version of Gleam releases.


lpil (approved):
Thank you!

PR #478, opened 2025-02-25
Updated fly deployment guide

metruzanca:
After following the guide, I noticed the fly deployment wasn't working and fly was complaining in the logs that the application was only listening on `127.0.0.1` instead of `0.0.0.0`.


metruzanca:
Realized, this is a duplicate of https://github.com/gleam-lang/website/pull/448

PR #482, opened 2025-03-09
Add echo example

GearsDatapacks:
As requested by Hayleigh on discord

### DIFF
--- a/_posts/2025-03-08-hello-echo-hello-git.md
+++ b/_posts/2025-03-08-hello-echo-hello-git.md
@@ -50,10 +50,8 @@ pub fn main() {
 
 The output of this program will look like this:
 
-```txt
-src/main.gleam:2
-[1, 2, 3]
-```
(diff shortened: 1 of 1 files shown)


lpil (commented):
Thank you. This is missing the colour of the output though!

GearsDatapacks:
Ah true. How can we add colour? Does the website support ansi highlighting?

Rhal95:
The website uses highlightjs. I don't think it supports exact custom highlighting but you can use one of the build in schemas.
You used `txt` (alias for `plaintext`) which effectively disables highlighting.
I pasted the example [here](https://highlightjs.org/demo) and it figured out that `apache` does some highlighting.
I can not try it out rn but I think it should be fine if you remove the `txt` in the code block to let the lib figure out the highlighting

GearsDatapacks:
Colours should now be added, if I've done it correctly

lpil (approved):
Thank you!!

Issue #479, opened 2025-02-26
Gleam guides


lpil:
## Databases

- [ ] Use Postgres with Pog

## HTTP

- [ ] Build an HTTP server using Wisp and Mist
- [ ] Configure TLS to serve HTTP with Mist
- [ ] Make HTTP requests on Erlang
- [ ] Make HTTP requests on JavaScript
- [ ] Make HTTP requests to JSON APIs on Erlang
- [ ] Make HTTP requests to JSON APIs on JavaScript

## Data encoding

- [ ] Compress and decompress data with DEFLATE
- [ ] Compress and decompress data with gzip
- [ ] Encode and decode base64 strings

## JSON

- [ ] Build JSON
- [ ] Decode JSON data

## Package management

- [ ] Add a dependency
- [ ] Add a git dependency
- [ ] Add a local path development 
- [ ] Use a custom CA certificate

## Operating system

- [ ] Listen to OS signals on Erlang
- [ ] Listen to OS signals on JavaScript
- [ ] Listen for CTRL+C on Erlang
- [ ] Listen for CTRL+C on JavaScript
- [ ] Read environment variables
- [ ] Write environment variables
- [ ] Read command-line arguments
- [ ] Read from stdin
- [ ] Run a Shell Command

## HTML

- [ ] Escape an HTML string
- [ ] Render HTML with Lustre (SSR)

## Time

- [ ] Sleep for a fixed number of milliseconds on Erlang
- [ ] Sleep for a fixed number of milliseconds on JavaScript
- [ ] Get the current local time
- [ ] Parse a RFC3339 timestamp string
- [ ] Convert a timestamp to a RFC3339 format string

## Security

- [ ] Hash a password
- [ ] Verify a checksum

## Working with files

- [ ] Append content to a file
- [ ] Check if a file exists
- [ ] Copy a file to another location
- [ ] Create a directory
- [ ] Create nested directories recursively
- [ ] Delete a file
- [ ] Delete a directory
- [ ] Get the MIME type of a file
- [ ] Get the current working directory
- [ ] Get the permissions of a file
- [ ] Get the size of a file
- [ ] Move a file
- [ ] Read a JSON file
- [ ] Read a binary file
- [ ] Read a text file
- [ ] Read the permissions of a file
- [ ] Set the permissions of a file
- [ ] Write JSON to a file
- [ ] Write a bit array to a file
- [ ] Write a string to a file

## Production

- [ ] Track exceptions locally
[...]

PR #488, opened 2025-03-13
Guides WIP/POC

nathanjcollins:
@lpil #479 I think is a great idea, and something I felt lacked while learning Gleam (still learning, so bear with me). I've done a couple POCs for guides, thinking that they should feel like "recipes", rather than fully fledged guides that explain the code line by line. What were your thoughts re. this? Should they be more verbose? Or should they be short and sweet (see my JSON example)?

### DIFF
--- a/guides.md
+++ b/guides.md
@@ -1,0 +1,22 @@
+---
+layout: page
+title: Guides
+subtitle: Learn from examples
+---
+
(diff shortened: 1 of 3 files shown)


lpil:
Thank you! We're not quite ready to accept contributions here as we're still working out how to implement this. Once we have the foundation in place there will be loads of guides to make

PR #490, opened 2025-04-04
Add pan flag Lucy

darenc:
Added a pan flag version of Lucy. Colour codes taken from https://www.flagcolorcodes.com/pansexual and https://www.sexualdiversity.org/edu/flags/1059.php

I've created this in the hope it will lead to a pan version of the Lucy stickers in the Gleam shop. =)


graphiteisaac:
Hey, would you like to add this to the [branding page](https://gleam.run/branding/) also? It looks That's where many of the Lucy variants are displayed.

darenc:
Good call, I've done that. Thanks for the suggestion.

I _haven't_ built the site locally to check it works as expected as I don't have Jekyll or Ruby installed. The HTML itself is straightforward, but I can install those things and build locally if you prefer.

lpil (approved):
Brilliant, thank you!

PR #495, opened 2025-05-01
Fix git SSH urls in gleam.toml example

sbergen:
They were using `git@github.com/...` when they should be `git@github.com:...`.
(Combined with https://github.com/gleam-lang/gleam/issues/4492 this has been causing confusion lately.)


lpil (approved):
Thank you

PR #496, opened 2025-05-04
Fixes: Removes mentions of fly.io legacy free allowance

hunterwilhelm:
Hello, this PR updates the deployment docs to remove mentions of the old free allowance, since [that plan is now legacy](https://fly.io/docs/about/pricing/#legacy-free-allowances).

### DIFF
--- a/src/website/page.gleam
+++ b/src/website/page.gleam
@@ -679,13 +679,8 @@ pub fn deployment_flyio(ctx: site.Context) -> fs.File {
     html.p([], [
       html.a([attr.href("https://fly.io")], [html.text("Fly.io")]),
       html.text(
-        " is a convenient and easy to use deployment platform with a generous ",
-      ),
-      html.a([attr.href("https://fly.io/docs/about/pricing/")], [
-        html.text("free allowance"),
-      ]),
-      html.text(
-        ". They were also a sponsor of the Gleam project, thank you Fly!",
+        " is a convenient and easy to use deployment platform. They were also a
+sponsor of the Gleam project, thank you Fly!",
       ),
     ]),
     html.h2([attr.id("prepare-your-application")], [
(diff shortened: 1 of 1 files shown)


lpil:
We were talking about the free allowance, not a plan. I'm not familiar with the old plan system.

Seems the allowance isn't documented on that page though... I'm not sure if that means we should be talking about it or now.

hunterwilhelm:
Keeping this might inadvertently mislead users, as it did in my case. 

The gleam docs currently state:
> [Fly.io](https://fly.io/) is a convenient and easy to use deployment platform with a generous [free allowance](https://fly.io/docs/about/pricing/).
> https://gleam.run/deployment/fly/

However, fly.io no longer appears to offer free allowances to new customers. I looked into this a bit and found:

### From the forum
Pay as You Go is the only option
> "New accounts now are all on the Pay as You Go system, which has no free allowances or trial credits." 
> https://community.fly.io/t/legacy-hobby/22218/2

They are temporarily waiving some invoices under $5 as a courtesy. But it might silently go away.
> "There is no free tier†, and the $5 waiver is just an informal courtesy—not a promise of eternal freebies. ... Personally, I wouldn’t be surprised if it was quietly changed in early 2026 (the second anniversary of the removal of the real free tier), to make it accumulate uncollected amounts."
> https://community.fly.io/t/concern-about-unexpected-charges-on-fly-io-free-tier-usage/24560/4


### From Fly.io’s pricing page
This shows the free allowance was part of the legacy hobby plan. The new Pay as You Go system doesn't mention any free usage.
> Screenshot 2025-05-07 at 5 34 29 PM
> https://fly.io/docs/about/pricing/#legacy-free-allowances

### From inside their website

At the top of the page it says: "Add a payment method to keep using our platform."
> Screenshot 2025-05-07 at 5 55 34 PM

I was hoping to use the free tier myself, which is what led me to dig into this. If you happen to come across any way it’s still available for new users, I’d definitely appreciate a heads-up.

lpil:
The $5 limit is the free allowance. I don't know about this other plan you're talking about here.

PR #498, opened 2025-05-19
Disambiguate 2nd paragraph under 'Reliable and scalable'

nnuuvv:
I was talking to other people about gleam and they got a little confused about gleams concurrency.

In their words:
> It's one of those new age erlang vm languages with "better concurrency"

After clearing the confusion they elaborated on what caused them to think that:

> I'd assume that if they were talking about the erlang vm, they'd say "its" or "BEAMs", instead of "a". 

So i swapped "a" to "it's"

### DIFF
--- a/src/website/page.gleam
+++ b/src/website/page.gleam
@@ -3508,7 +3508,7 @@ pub fn main() {
           ]),
           html.p([], [
             html.text(
-              "Thanks to a multi-core actor based concurrency system that can run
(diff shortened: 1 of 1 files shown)


lpil:
I think this should be "its" rather than "it's"?

nnuuvv:
yes, yes it should be

lpil (approved):
Thank you!

PR #504, opened 2025-05-29
Correct Zed LSP steps

strawmelonjuice:
It seems the Zed Editor has changed Gleam support to now be in a plugin, showing this prompt on first opening a Gleam file:

Zoom in to prompt

This installs the following plugin:
plugin entry in search

And only after downloading that plugin the LSP works.

So I reflected that change in this PR :)


lpil (approved):
Thank you!

PR #507, opened 2025-06-04
fix #506 - mobile issue caused by lambdaclass logo

### LINKED ISSUE #506: fix small issues in the ui
Image

☝️the logo is pushing is going outside the container  like shown below 

https://gleam.run/news/gleam-javascript-gets-30-percent-faster/

Image

the same thing happens here there is small gap caused by this element 
https://gleam.run/

Image

graphiteisaac:


lpil (approved):
Thank you

PR #512, opened 2025-06-30
Centre hero text when not on the homepage

graphiteisaac:
Hey, short and simple one today to fix some awkward styling across the website. Currently, we always left-align text in the hero (header), which leads to some headings looking a bit off, like this:
image

This is correct for the homepage though, because it's balanced against Lucy on the left side, so we need a class (text-left) to differentiate it. I've decided to go for a reusable utility class rather than a single-use "hero-home" class or similar.

After this change, here's how page hero text should look:
image

### DIFF
--- a/src/website/page.gleam
+++ b/src/website/page.gleam
@@ -3487,7 +3487,7 @@ pub fn home(ctx: site.Context) -> fs.File {
         "Lucy the star, Gleam's mascot",
       )),
       content: [
-        html.div([], [
+        html.div([class("text-left")], [
           html.b([], [html.text("Gleam")]),
           html.text(" is a "),
           html.b([], [html.text("friendly")]),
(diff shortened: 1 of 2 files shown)


lpil:
We don't do merge commits! Can you rebase your branch to remove them please 🙏

src/website/page.gleam
```diff
@@ -3487,7 +3487,7 @@ pub fn home(ctx: site.Context) -> fs.File {
         "Lucy the star, Gleam's mascot",
       )),
       content: [
-        html.div([], [
+        html.div([class("text-left")], [
```
lpil: I think the alignment should be done in the `header` function, setting it to centre when `hero_image` is `option.None`. Otherwise we'd get text that is centred even when a image is present.

PR #513, opened 2025-07-10
New maple Lucy svg image

toulerl:
I have added Maple Lucy. Have a look if that one is ok  :)

maple lucy


lpil:
Could you add it to the branding page also please 🙏

toulerl:
> Could you add it to the branding page also please 🙏

I completely missed that page, yes I will add

PR #514, opened 2025-07-11
Add Strand case study

lpil:


graphiteisaac:
Should the updates to sponsor.gleam be included in this PR? Seems unrelated, from what I can tell

lpil:
Eh it's an automated update 🤷

graphiteisaac:
Fixed formatting for the CSS I keep forgetting to turn it off, woops!

PR #515, opened 2025-07-13
Use created user in deployment dockerfile

JonasGruenwald:
The `webapp` user is added but not used, in the docs for deployment on a Linux server


lpil (approved):
Thank you!

PR #525, opened 2025-08-05
fixing a few typos

rawhat:


posts/no-more-dependency-management-headaches.djot
```diff
@@ -96,7 +96,7 @@ lustre     3.1.4 -> 5.1.1
 
 Thank you [Amjad Mohamed](https://github.com/andho) for this!
 
-## Future dependency management improvements
```
rawhat, line 99: oops sorry, i have a "remove whitespace" formatter.  hopefully these aren't too annoying

lpil: Nah that's good. I need to make my editor do that.

PR #527, opened 2025-08-08
Dedicated Sponsor Page

graphiteisaac:
Hello (again) Gleam gang! After getting some feedback from Louis, I've made some changes to my sponsor page PR (and restarted it, to fix the issue of going off of `main`)

Here's a screenshot, for a quick and easy way to check it out.
image

Same as last time, there are some things worth considering before merging in, namely the specifics of the prose & whether everyone is happy with their avatars & titles.

### DIFF
--- a/src/website/page.gleam
+++ b/src/website/page.gleam
@@ -166,8 +166,7 @@ pub fn sponsor(ctx: site.Context) -> fs.File {
               "Even if it’s just $1/month, it’s one dollar more towards the development of Gleam! Also, the fact that you’re sponsoring Gleam is visible on your GitHub profile, giving the language additional exposure and serving as a signal that you care about it and might encourage others to sponsor Gleam as well 🩷",
             ),
           ]),
-        ]),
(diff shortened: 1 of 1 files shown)


lpil:
# Prioritising conversion

It's quite a long way before there's a call-to-action, and when there is one it's directing people to read the roadmap, **not to sponsor**.

The goal is to get people to sponsor, so I think we should focus on that as much as possible.

## Current design

image


## With the call to action prominently on the page

image


Quick and dirty version, to give you an idea of an alternative.

Here the first thing you are presented with is the options to sponsor folks, and the above the fold content is more visually engaging.

The rest of the page would need some rework, and the copy doesn't flow quite so well any more.

# Differentiate sponsoring the lpil account

Let's have it say "Sponsor the project" or something for my github account, and "Sponsor $name" for the others.

# Mobile

These cards don't work so well on mobile

image

Kamila-P:
new copy proposal updated in the original doc

PR #530, opened 2025-08-14
Add a fistful of missing spaces after links

frunobulax-the-poodle:
I found a few missing spaces on the FAQ page and then poked around some more, but it's probably not exhaustive.


graphiteisaac:
Really appreciate this! There are a few spots like this on the site but this seems to get a great chunk of them out of the way.

Issue #521, opened 2025-08-03
Share button seems to do nothing


lpil:
The UX is a bit confusing currently at it doesn't inform the user that anything has happened, making it seem broken.

I wonder if we should remove the share button as all it does currently is copy the URL, but browsers already have a convenient way to do that, and the visitor will be more familiar with that.

PR #534, opened 2025-08-27
Fix tooltip not existing

graphiteisaac:
Somewhere along my merry way I lost the tooltip for the share button on the news page, so here it is patched. This is intended to be modular enough that we can use it in other places on the site.

Addresses #521

### DIFF
--- a/src/website/page.gleam
+++ b/src/website/page.gleam
@@ -397,7 +397,7 @@ pub fn news_post(post: news.NewsPost, ctx: site.Context) -> fs.File {
             ),
             attr.data("tooltip-position", "right"),
             attr.data("tooltip-trigger", "click"),
-            attr.aria_describedby("share-tooltip"),
+            attr.aria_label("Copy post URL to clipboard"),
             class("tooltip-container meta-button share-button"),
           ],
           [
@@ -412,8 +412,7 @@ pub fn news_post(post: news.NewsPost, ctx: site.Context) -> fs.File {
                 class("tooltip"),
                 attr.id("share-tooltip"),
                 attr.role("status"),
-                attr.aria_atomic(True),
-                attr.aria_live("polite"),
+                attr.aria_hidden(True),
               ],
               [html.text("Copied the post URL!")],
             ),


src/website/page.gleam
```diff
@@ -395,6 +395,9 @@ pub fn news_post(post: news.NewsPost, ctx: site.Context) -> fs.File {
                 <> post.path
                 <> "')",
             ),
+            attr.data("tooltip", "Copied the post URL!"),
+            attr.data("tooltip-position", "right"),
+            attr.data("tooltip-toggle", "click"),
```
lpil: It says toggle but it looks more like a trigger.

Could get rid of this? idk if there's triggers other than click?

lpil (commented):
Oh nice! I like the way it works. Left a Q inline.

Are there aria things to add?

src/website/page.gleam
graphiteisaac: The reason for this is to make it clear that we can implement different kinds of tooltips, I have hover-activated ones in mind for other pages. If you think it's a bit premature to do this I'm happy to simplify it

graphiteisaac:
> Are there aria things to add?

Very good point. I'm going to change how this is implemented to better support screen readers because it turns out they can't reference pseudo-elements.

src/website/page.gleam
```diff
@@ -395,7 +395,10 @@ pub fn news_post(post: news.NewsPost, ctx: site.Context) -> fs.File {
                 <> post.path
                 <> "')",
             ),
-            class("meta-button share-button"),
+            attr.data("tooltip-position", "right"),
+            attr.data("tooltip-trigger", "click"),
+            attr.aria_describedby("share-tooltip"),
```
hayleigh-dot-dev: I don't think describedby is the correct attribute in this case. As it stands, when a user tabs to this button a screen reader will announce: "Share. Copied the post URL!".

Instead, remove the relationship between the button and the live region, initially render no text content in the live region, and on click set the live region's text content to "Copied the post URL."

hayleigh-dot-dev (changes requested):

src/website/page.gleam
hayleigh-dot-dev: Alternatively, because aria stuff is hard and bad aria is much worse than no aria, we could consider hiding the tooltip from the accessibility tree entirely and putting `attriburte.aria_label("Copy post URL to clipboard")` on the button. The user doesn't need to be notified that the action happened really.

graphiteisaac: I think I semantically prefer option 2? It makes more sense to alert people to what the button does if they aren't seeing it. Does that seem right to you?

hayleigh-dot-dev: Yep sounds good to me, little reason to alert people that "the thing we said would happen did happen" in this case!

hayleigh-dot-dev: For explicitness-sake, that means

- `aria-label` on the button to announce what it does.
- `aria-hidden` on the tooltip to hide it from the accessibility tree.

hayleigh-dot-dev (approved):

lpil (approved):
Thank you!!

PR #535, opened 2025-08-29
Improve case study styling

graphiteisaac:
G'day there Gleamers

Simple one today, to address the case study page looking a bit bland and boring, I've added a little info box to the side and a quote above (Thanks Hayleigh!) 

image


lpil (commented):
Isaac, you're my hero

PR #536, opened 2025-08-29
fix share button for case studies & fix some styling duplication

graphiteisaac:
Very silly mistake I've made here. The share button was only updated for the news pages, not the case study.

Have also removed some duplicated styling.


src/website/page.gleam
```diff
@@ -353,19 +353,31 @@ pub fn case_study(post: case_study.CaseStudy, ctx: site.Context) -> fs.File {
           [
             attr(
               "onclick",
-              "window.navigator.clipboard.writeText('https://gleam.run/case-studies/"
+              "window.navigator.clipboard.writeText('https://gleam.run/news/"
```
lpil: Could make it copy the current URL to avoid this problem in future?

PR #537, opened 2025-09-01
Add the open-quote on case-study view mobile

graphiteisaac:
Bit of an oversight, seems I was looking at this long enough I didn't really see how odd it looked to not have the quotation mark on mobile. Alas, she returns:

image


lpil (approved):
Thank you!!!

PR #543, opened 2025-09-18
fix: add missing spaces to cheatsheet

yagogarea:
Fixed missing spaces in `cheatsheet` that caused highlighted names to appear stuck to the surrounding text on the web.


lpil (approved):
Thank you!!

PR #552, opened 2025-10-06
Table of contents layout

graphiteisaac:
G'day!

Today I come bearing a new page layout, the table of contents layout! I've had this one in the pipelines for a bit now, but I've properly factored it as a separate unit so it can be used now. 

This includes a small refactor, adding a new layout.gleam file with our basic layout functions, giving some clearer distinction between page content and layout things. It also includes a function that parses Djot document content to produce a List of ContentLinks, used to build the table of contents in the layout.

For now, since I want to start small(er), I have just set it up manually on the two deployment pages. 

image

### DIFF
--- a/src/website/cheatsheet.gleam
+++ b/src/website/cheatsheet.gleam
@@ -2,13 +2,12 @@ import gleam/option
 import lustre/attribute.{attribute as attr} as attr
 import lustre/element/html
 import website/fs
-import website/layout
 import website/page
 import website/site
 
 pub fn elixir(ctx: site.Context) -> fs.File {
   let meta =
-    site.PageMeta(
+    page.PageMeta(
       path: "cheatsheets/gleam-for-elixir-users",
       title: "Gleam for Elixir users",
       meta_title: "Gleam for Elixir users | Cheat sheet",
@@ -1298,13 +1297,13 @@ let assert #(1 as a, 2 as b) = #(1, 2)
       ]),
     ]),
   ]
-  |> layout.with_header("roadmap", meta, ctx)
+  |> page.layout_header("roadmap", meta, ctx)
   |> page.to_html_file(meta)
 }
 
 pub fn erlang(ctx: site.Context) -> fs.File {
   let meta =
-    site.PageMeta(
+    page.PageMeta(
       path: "cheatsheets/gleam-for-erlang-users",
       title: "Gleam for Erlang users",
       meta_title: "Gleam for Erlang users | Cheat sheet",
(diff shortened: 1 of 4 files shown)


src/website/layout.gleam
```diff
@@ -0,0 +1,292 @@
...
+import website/site
+
+pub type ContentLink {
+  ContentLink(title: String, href: String, children: List(ContentLink))
+}
+
+pub fn table_of_contents_from_djot(document: jot.Document) -> List(ContentLink) {
+  document.content
+  |> list.fold(from: [], over: _, with: fn(accum, block) {
+    case block {
```
lpil: Flatten these nested case expressions and the `list.first` into a single case expression please.

src/website/layout.gleam
```diff
@@ -0,0 +1,292 @@
...
+      jot.Heading(attributes:, level:, content:) -> {
+        let href =
+          "#"
+          <> dict.get(attributes, "id")
+          |> result.unwrap("")
```
lpil: Skip if id is missing instead of rendering an invalid link

src/website/layout.gleam
```diff
@@ -0,0 +1,292 @@
...
+      [],
+      list.append(page_content, [
+        footer(ctx),
+        html.script([attr.src("/javascript/main.js"), attr("async", "")], ""),
+      ]),
+    ),
+  ])
+}
+
+pub fn with_toc(
```
lpil: No abbreviations in Gleam code please!

src/website/layout.gleam
```diff
@@ -0,0 +1,292 @@
+import gleam/dict
```
lpil: Merge this module back into the page module please.

The fact the `PagaMeta` data type no longer has an appropriate to live without creating dependency cycles shows that this is one module and not two.

src/website/layout.gleam
```diff
@@ -0,0 +1,292 @@
...
+import lustre/attribute.{attribute as attr, class} as attr
+import lustre/element.{type Element}
+import lustre/element/html
+import website/site
+
+pub type ContentLink {
+  ContentLink(title: String, href: String, children: List(ContentLink))
+}
+
+pub fn table_of_contents_from_djot(document: jot.Document) -> List(ContentLink) {
```
lpil: Is this used anywhere?

src/website/site.gleam
```diff
@@ -1,9 +1,24 @@
...
+    title: String,
+    subtitle: String,
+    meta_title: String,
+    description: String,
+    /// Render-critical to pre-load using meta-tags
+    preview_image: Option(String),
+    /// Social media share preview image name
+    preload_images: List(String),
+  )
+}
```
lpil: This belongs in the page module!

lpil (commented):
Thank you!! I've left a few lil notes

src/website/layout.gleam
graphiteisaac: It is not currently, but it will be as soon as you use djot to write a guide like this.

graphiteisaac:
Thanks, I've implemented much of this feedback now.

src/website/layout.gleam
lpil: ooo smart

PR #561, opened 2025-10-20
Rewrite heading in Gleam v1.13.0 release

mtlynch:
The current phrasing makes it sound like the release makes warnings 'more inefficient', which I don't think it the intent.

Suggest this rephrasing to make the intent more clear.


lpil (approved):
Thanks!

PR #563, opened 2025-10-20
use tear to highlight elixir

graphiteisaac:
and also fix elixir map syntax being incorrect (using #{} instead of %{})

Thanks to @yoshie-monster for this package!


yoshi-monster:
I noticed yesterday that had the exact same bitarray highlighting bug that contour did, so it would be great to update both ^.^

~ 💜

lpil (approved):
Amazing!!!!!!

PR #564, opened 2025-10-21
Improve syntax highlighting

yoshi-monster:
`tear` uses some additional classes for highlighting, so I added selectors to existing styles to make it look consistent with Erlang/Gleam.

I also updated tear and contour to fix the highlighting of bit arrays.

### DIFF
--- a/priv/styles/main.css
+++ b/priv/styles/main.css
@@ -751,6 +751,7 @@ main.content {
 .hl-module,
 .hl-variant,
 .hl-symbol,
+.hl-atom,
 .hl-literal,
 .code-type,
 .code-module {


priv/styles/main.css
```diff
@@ -729,6 +729,7 @@ main.content {
 
 .code-warning,
 .hl-keyword,
+.hl-built_in,
```
lpil, line 732: How come this is a mix of snake and kebab case? Shouldn't it be `hl-built-in`, as CSS uses kebab case?

yoshi-monster: I kept the `hl` prefix, the class name `built_in` is taken from highlight.js.
I'm not very fond of it either, but yeah that's where it comes from :D

I might have been just too tired yesterday when I added the html formatter

lpil: Blimey. What is hljs up to

PR #595, opened 2026-01-30
Update FAQ with clarifications on Elixir vs Gleam

josevalim:
This is mostly for discussion, I am adding comments with more context.

### DIFF
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -151,7 +151,8 @@ Here’s a non-exhaustive list of differences:
 - Elixir’s compiler is written in Erlang and Elixir, Gleam’s is written in Rust.
 - Gleam has a more traditional C family style syntax.
 - Elixir has a namespace for module functions and another for variables, Gleam
-  has one unified namespace (so there’s no special fun.()syntax).
+  has one unified namespace (there is no `fun.()` syntax in Gleam, but in Gleam
+  variable names can override function names).
 - Gleam standard library is distributed as Hex packages, which makes
   interoperability with other BEAM languages easier.
 - Elixir is a larger language, featuring numerous language features not present in Gleam.
@@ -164,18 +165,15 @@ Here’s a non-exhaustive list of differences:
 - Elixir has superior BEAM runtime integration, featuring accurate stack traces
   and full support for tools such as code coverage, profiling, and more.
   Gleam’s support is much weaker due to going via Erlang source.
-- Elixir and Gleam both use Erlang's OTP framework. Both have additional
-  modules for working with OTP, which provide APIs more in the style of each
-  respective language. Both common use Erlang's OTP APIs directly, but Elixir
-  can do so more conveniently and concisely due to having a less-strict type
-  system.
+- Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
+  modules directly and also includes its own abstractions (such as tasks).
+  Gleam provide APIs more in the style of each respective language.
 - Elixir currently has superior deployment tooling, including support for OTP
   releases and OTP umbrella applications.
 - Gleam’s editor tooling is superior due to having a more mature official
   language server, but Elixir has recently announced an official language
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
-- Gleam compiles faster than Elixir.
 
 Most importantly both are BEAM languages! I advise using whichever has the
 programming style you personally find most enjoyable and productive.
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -152,7 +152,7 @@ Here’s a non-exhaustive list of differences:
 - Gleam has a more traditional C family style syntax.
 - Elixir has a namespace for module functions and another for variables, Gleam
   has one unified namespace (there is no `fun.()` syntax in Gleam, but in Gleam
-  variable names can override function names).
+  variable names can shadow function names).
 - Gleam standard library is distributed as Hex packages, which makes
   interoperability with other BEAM languages easier.
 - Elixir is a larger language, featuring numerous language features not present in Gleam.
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -179,7 +179,7 @@ Here’s a non-exhaustive list of differences:
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
 - Gleam compiles faster than Elixir (measured by the time taken to
-  compiling a project with 100 single-function files in each language)
+  compiling a project with 100 single-function files in each language).
 
 Most importantly both are BEAM languages! I advise using whichever has the
 programming style you personally find most enjoyable and productive.
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -165,9 +165,6 @@ Here’s a non-exhaustive list of differences:
 - Elixir has superior BEAM runtime integration, featuring accurate stack traces
   and full support for tools such as code coverage, profiling, and more.
   Gleam’s support is much weaker due to going via Erlang source.
-- Elixir documentation adheres to the official Erlang VM format,
-  which facilitates interoperability between documentation tooling.
-  Gleam does not store documentation metadata in source files.
 - Elixir interactive and remote shells support writing both Elixir
   and Erlang code. Gleam shells require writing Erlang code.
 - Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -167,7 +167,7 @@ Here’s a non-exhaustive list of differences:
   Gleam’s support is much weaker due to going via Erlang source.
 - Elixir documentation adheres to the official Erlang VM format,
   which facilitates interoperability between documentation tooling.
-  Gleam does not store documentation metadata in source files
+  Gleam does not store documentation metadata in source files.
 - Elixir interactive and remote shells support writing both Elixir
   and Erlang code. Gleam shells require writing Erlang code.
 - Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -165,6 +165,11 @@ Here’s a non-exhaustive list of differences:
 - Elixir has superior BEAM runtime integration, featuring accurate stack traces
   and full support for tools such as code coverage, profiling, and more.
   Gleam’s support is much weaker due to going via Erlang source.
+- Elixir documentation adheres to the official Erlang VM format,
+  which facilitates interoperability between documentation tooling.
+  Gleam does not store documentation metadata in source files
+- Elixir interactive and remote shells support writing both Elixir
+  and Erlang code. Gleam shells require writing Erlang code.
 - Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
   modules directly and also includes its own abstractions (such as tasks).
   Gleam provide APIs more in the style of each respective language.
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -178,8 +178,6 @@ Here’s a non-exhaustive list of differences:
   language server, but Elixir has recently announced an official language
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
-- Gleam compiles faster than Elixir (measured by the time taken to
-  compiling a project with 100 single-function files in each language).
 
 Most importantly both are BEAM languages! I advise using whichever has the
 programming style you personally find most enjoyable and productive.
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -167,15 +167,19 @@ Here’s a non-exhaustive list of differences:
   Gleam’s support is much weaker due to going via Erlang source.
 - Elixir interactive and remote shells support writing both Elixir
   and Erlang code. Gleam shells require writing Erlang code.
-- Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
-  modules directly and also includes its own abstractions (such as tasks).
-  Gleam provide APIs more in the style of each respective language.
+- Elixir and Gleam both use Erlang's OTP framework. Both languages can
+  use Erlang/OTP modules directly, though it is more ergonomic and
+  requires less boilerplate in Elixir. Both languages offer their own
+  additional abstractions, with Elixir providing new functionality
+  and Gleam providing type-safe interfaces.
 - Elixir currently has superior deployment tooling, including support for OTP
   releases and OTP umbrella applications.
 - Gleam’s editor tooling is superior due to having a more mature official
   language server, but Elixir has recently announced an official language
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
+- Gleam compiles faster than Elixir (measured by the time taken to
+  compiling a project with 100 single-function files in each language)
 
 Most importantly both are BEAM languages! I advise using whichever has the
 programming style you personally find most enjoyable and productive.


documentation/frequently-asked-questions.djot
```diff
@@ -151,7 +151,8 @@ Here’s a non-exhaustive list of differences:
 - Elixir’s compiler is written in Erlang and Elixir, Gleam’s is written in Rust.
 - Gleam has a more traditional C family style syntax.
 - Elixir has a namespace for module functions and another for variables, Gleam
-  has one unified namespace (so there’s no special fun.()syntax).
+  has one unified namespace (there is no `fun.()` syntax in Gleam, but in Gleam
+  variable names can override function names).
```
josevalim: The single namespace has two downsides, which were not listed:

1. It allows variable names to override function names (i.e. it introduces ambiguity)
2. It makes interoperability with Erlang harder (as it also two namespaces semantically)

So I decided to add one of the downsides. But I'd say this distinction is very minor, I'd nuke it.

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,15 @@ Here’s a non-exhaustive list of differences:
...
   and full support for tools such as code coverage, profiling, and more.
   Gleam’s support is much weaker due to going via Erlang source.
-- Elixir and Gleam both use Erlang's OTP framework. Both have additional
-  modules for working with OTP, which provide APIs more in the style of each
-  respective language. Both common use Erlang's OTP APIs directly, but Elixir
-  can do so more conveniently and concisely due to having a less-strict type
-  system.
+- Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
+  modules directly and also includes its own abstractions (such as tasks).
+  Gleam provide APIs more in the style of each respective language.
```
josevalim: I have rephrased the Elixir side of things. The Gleam side of things is now incomplete, but I don't feel competent to describe it. Help is appreciated.

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,15 @@ Here’s a non-exhaustive list of differences:
...
 - Elixir currently has superior deployment tooling, including support for OTP
   releases and OTP umbrella applications.
 - Gleam’s editor tooling is superior due to having a more mature official
   language server, but Elixir has recently announced an official language
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
-- Gleam compiles faster than Elixir.
```
josevalim, line 178: I removed this because I am not aware of any data to support this but perhaps you are aware of some? There is an argument that Gleam compiles faster, because the Gleam compiler emits Erlang source files, but I don't think that's an apples to apples comparison, because when you run a command such as `gleam test`, it would still require those files to be compiled into .beam files. A good test would be the time to compile and start an app that immediately shuts down.

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,15 @@ Here’s a non-exhaustive list of differences:
 - Elixir has superior BEAM runtime integration, featuring accurate stack traces
   and full support for tools such as code coverage, profiling, and more.
   Gleam’s support is much weaker due to going via Erlang source.
```
josevalim, line 167:
```suggestion
  Gleam’s support is much weaker due to going via Erlang source.
- Elixir documentation adheres to the official Erlang VM format,
  which facilitates interoperability between documentation tooling.
  Gleam does not store documentation metadata in source files
- Elixir interactive and remote shells support writing both Elixir
  and Erlang code. Gleam shells require writing Erlang code.
```

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,20 @@ Here’s a non-exhaustive list of differences:
...
-- Elixir and Gleam both use Erlang's OTP framework. Both have additional
-  modules for working with OTP, which provide APIs more in the style of each
-  respective language. Both common use Erlang's OTP APIs directly, but Elixir
-  can do so more conveniently and concisely due to having a less-strict type
-  system.
+- Elixir documentation adheres to the official Erlang VM format,
+  which facilitates interoperability between documentation tooling.
+  Gleam does not store documentation metadata in source files
```
```suggestion
  Gleam does not store documentation metadata in source files.
```

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,20 @@ Here’s a non-exhaustive list of differences:
...
   and full support for tools such as code coverage, profiling, and more.
   Gleam’s support is much weaker due to going via Erlang source.
-- Elixir and Gleam both use Erlang's OTP framework. Both have additional
-  modules for working with OTP, which provide APIs more in the style of each
-  respective language. Both common use Erlang's OTP APIs directly, but Elixir
-  can do so more conveniently and concisely due to having a less-strict type
-  system.
+- Elixir documentation adheres to the official Erlang VM format,
+  which facilitates interoperability between documentation tooling.
+  Gleam does not store documentation metadata in source files.
```
```suggestion
  Gleam does not store documentation metadata in source files.
```

You may want to add a "yet" here? Because now you could emit `-moduledoc(...)` and `-doc(...)` nodes, right?

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,20 @@ Here’s a non-exhaustive list of differences:
...
+- Elixir interactive and remote shells support writing both Elixir
+  and Erlang code. Gleam shells require writing Erlang code.
```
josevalim, line 169:
```suggestion
  and Erlang code. Gleam shells require writing Erlang (or JavaScript).
```

?

documentation/frequently-asked-questions.djot
josevalim: To give more context, an end to end compilation + test in Gleam is (I believe):

    boot up rust -> parse gleam -> write .erl -> boot up erlc -> parse .erl -> write .beam -> boot up erl -> start tests

And end to end compilation in Elixir is:

    boot up erlang -> parse elixir -> expand macros (may load a temporary .beam in memory) -> write .beam -> start tests

documentation/frequently-asked-questions.djot
giacomocavalieri: Hello Jose! Gleam already generates those `-moduledoc` and `-doc`, the feature was added back in February with the [v1.8.0 release](https://gleam.run/news/gleam-gets-rename-variable/#:~:text=Erlang%20documentation%20integration)

documentation/frequently-asked-questions.djot
lpil: Gleam doesn't have overriding, but it does have shadowing.

documentation/frequently-asked-questions.djot
lpil: This is the same for Gleam and for Elixir!

documentation/frequently-asked-questions.djot
josevalim:
```suggestion
  variable names can shadow function names).
```

documentation/frequently-asked-questions.djot
josevalim: I may be missing something, but you cannot use `gen_fsm` in Gleam directly, right? So in Gleam you need to add the typed interface on top?

documentation/frequently-asked-questions.djot
josevalim: Awesome job! I am removing this then!

```suggestion
```

documentation/frequently-asked-questions.djot
lpil: This is including the Erlang -> BEAM compilation too. Compiling Elixir is more challenging than compiling Gleam as Gleam is a simpler language, and Gleam's compiler performs faster than Elixir's due to using native code rather than the BEAM, so in like-for-like codebases Gleam consistently compiles faster.

Gleam also doesn't permit cycles between modules, so it has an easier job of avoiding slowdown in large codebases.

> I removed this because I am not aware of any data to support this but perhaps you are aware of some? 

I've measured! Historically I measured like-for-like, but in recent years I've only measured the diff between new and old versions of the same language to see how they have changed with each release.

If you have some newer data that shows things have changed then we can correct this to show the current state of play.

documentation/frequently-asked-questions.djot
lpil: We don't have any tooling for shells in JavaScript today, just the Erlang one.

documentation/frequently-asked-questions.djot
lpil: You can use any untyped API in Gleam, people just tend to use typed variants as they're nicer to work with.

documentation/frequently-asked-questions.djot
```diff
@@ -151,7 +151,8 @@ Here’s a non-exhaustive list of differences:
 - Elixir’s compiler is written in Erlang and Elixir, Gleam’s is written in Rust.
 - Gleam has a more traditional C family style syntax.
 - Elixir has a namespace for module functions and another for variables, Gleam
-  has one unified namespace (so there’s no special fun.()syntax).
+  has one unified namespace (there is no `fun.()` syntax in Gleam, but in Gleam
```
lpil, line 154: oops, thanks for fixing the syntax error here too

documentation/frequently-asked-questions.djot
josevalim: In order to be a bit more concrete, perhaps a suggestion is:

```suggestion
- Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
  modules directly and also includes its own abstractions (such as tasks).
  Gleam provides its own type-safe abstractions and requires declaring foreign
  function interfaces for existing Erlang/OTP modules.
```

But I am sure you have a better way to phrase the above. :)

lpil: Perhaps it would be good to say something like this:

> - Elixir and Gleam both use Erlang's OTP framework. Both languages can
>  use Erlang/OTP modules directly, though it is more ergonomic and
>  requires less boilerplate in Elixir. Both languages offer their own
>  additional abstractions, with Elixir providing new functionality
> and Gleam providing type-safe interfaces.

documentation/frequently-asked-questions.djot
josevalim: Ok, that gives me a reference point! I asked Claude to create two projects with 100 single-function files, ran `gleam test` vs `mix test` in each and I got 0.5s for Gleam, 0.7s for Elixir, so I am more than happy to add it back!

I will try to add a reference point, if that's ok.

documentation/frequently-asked-questions.djot
josevalim: This is excellent. Let me add that back in.

```suggestion
- Elixir and Gleam both use Erlang's OTP framework. Both languages can
  use Erlang/OTP modules directly, though it is more ergonomic and
  requires less boilerplate in Elixir. Both languages offer their own
  additional abstractions, with Elixir providing new functionality
  and Gleam providing type-safe interfaces.
```

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,17 @@ Here’s a non-exhaustive list of differences:
...
+  and Erlang code. Gleam shells require writing Erlang code.
+- Elixir and Gleam both use Erlang's OTP framework. Elixir can use Erlang/OTP
+  modules directly and also includes its own abstractions (such as tasks).
+  Gleam provide APIs more in the style of each respective language.
 - Elixir currently has superior deployment tooling, including support for OTP
   releases and OTP umbrella applications.
 - Gleam’s editor tooling is superior due to having a more mature official
   language server, but Elixir has recently announced an official language
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
```
josevalim, line 180:
```suggestion
- Elixir is more mature than Gleam and has a much larger ecosystem.
- Gleam compiles faster than Elixir (measured by the time taken to
  compiling a project with 100 single-function files in each language)
```

I add some context some anyone can come, run the experiments, and validate the results (and to avoid people like me who would be like "what is the data/reference?!" :P).

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,21 @@ Here’s a non-exhaustive list of differences:
...
+  and Gleam providing type-safe interfaces.
 - Elixir currently has superior deployment tooling, including support for OTP
   releases and OTP umbrella applications.
 - Gleam’s editor tooling is superior due to having a more mature official
   language server, but Elixir has recently announced an official language
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
-- Gleam compiles faster than Elixir.
+- Gleam compiles faster than Elixir (measured by the time taken to
+  compiling a project with 100 single-function files in each language)
```
```suggestion
  compiling a project with 100 single-function files in each language).
```

josevalim:
Thank you @giacomocavalieri and @lpil! I have answered all of my own questions, I am happy with the above, but if you want me to change anything, just let me know!

(it is also very cool you support `-moduledoc`/`-doc`, that made my day!)

documentation/frequently-asked-questions.djot
josevalim: I have done further work and benchmarks on this.

The reason why the Elixir compiler was slower in my benchmarks was because our macro expansion uses intermediate modules behind the scenes and we were deleting and purging those modules immediately after compilation. [Elixir v1.19 had already postponed the purging of modules](https://github.com/elixir-lang/elixir/blob/v1.19/CHANGELOG.md#code-loading-bottlenecks), because it goes through Erlang's code server, which is a single process and makes it a bottleneck. But deleting modules also goes through this same process and on my machine, with ten cores, that was showing up. [This commit I just pushed to main also postpones it like we did with purging](https://github.com/elixir-lang/elixir/commit/c40c140e3f524fa8384efd33137ec1c3fab341d3) and now, for the scenario above, Elixir compiles as fast as Gleam.

> Gleam also doesn't permit cycles between modules, so it has an easier job of avoiding slowdown in large codebases.

Elixir doesn't allow cycles between modules at compile time either. We do allow it at runtime, but that has no effect at compile time!

> Gleam's compiler performs faster than Elixir's due to using native code rather than the BEAM, so in like-for-like codebases Gleam consistently compiles faster.

In the last several years optimizing the Elixir compiler, the compiler front-end was almost never the bottleneck, only the compiler backend (i.e. the Erlang compiler), which in this case is shared between Gleam and Elixir. I agree that native code will be faster but Gleam's architecture also requires invoking `erlc` and starting `erl` separately.

Therefore, instead of guessing, I decided to run further benchmarks. This time, instead of giving single-function modules, each module has 100 functions. This means more work for both compiler frontends, which would give more opportunity for Gleam's `native` frontend to get ahead of Elixir's, but the times were the same for both on my machine: ~0.8s.

I have published the code here: https://github.com/josevalim/langcompilebench. I have measured using Elixir main (soon to be v1.20-rc.2) and Gleam v1.14.0.

Edit: oh, it is also worth saying that Elixir v1.19 made some codebases compile 3x faster. So I don't doubt Gleam codebases were consistently faster in the past, but there has been a lot of progress on our side here. :)

documentation/frequently-asked-questions.djot
```diff
@@ -164,18 +165,21 @@ Here’s a non-exhaustive list of differences:
...
+  and Gleam providing type-safe interfaces.
 - Elixir currently has superior deployment tooling, including support for OTP
   releases and OTP umbrella applications.
 - Gleam’s editor tooling is superior due to having a more mature official
   language server, but Elixir has recently announced an official language
   server project which is in active development.
 - Elixir is more mature than Gleam and has a much larger ecosystem.
-- Gleam compiles faster than Elixir.
+- Gleam compiles faster than Elixir (measured by the time taken to
+  compiling a project with 100 single-function files in each language).
```
```suggestion
```

documentation/frequently-asked-questions.djot
josevalim: More updates! I was able to speed up loading modules altogether and I have sent a potential pull request to Erlang/OTP: https://github.com/erlang/otp/pull/10615. The good news is that, while it improves the Elixir compiler, it also improves booting in general, which will benefit commands that run `erl` too, such as `gleam test`. I have tested a few Elixir projects and I saw a ~10% improvement on boot.

lpil: Great to hear about those improvements!

RE the benchmarks, I don't think that's a meaningful representation of code in either language. This is a description of the properties of typical real-world code, and many functions that do nothing other than return a literal.

As you say, the Elixir frontend has no issue. The difference between Gleam and Elixir is the way they use the backend, largely due to macros. Elixir generates more code and will compile multiple times to resolve macros, so the cost of the backend is more impactful. The module cycles compound the problem as they force more cache invalidation.

Let's undo this change in this PR so we can merge the other changes until there's new benchmarks.

josevalim: Can you please define then what is a meaningful benchmark? You have said that:

> If you have some newer data that shows things have changed then we can correct this to show the current state of play.

But without an actual reference or example code of what it is meaningful, then it is really hard to draw any conclusions, and, at least for Elixir v1.20, the blank claim that Gleam is faster will not hold for boilerplate code (which I agree is a poor example, but it is at least an example).

> The module cycles compound the problem as they force more cache invalidation.

I am not sure what this means. Can you please provide more context and how we could measure said cache invalidation?

All I am saying is, I’d love to look at some concrete data, because it means we can all collectively improve. And if the issues are rooted in Erlang, then I can improve the Erlang compiler, and that’s a win-win for everyone.

josevalim: > I am not sure what this means. Can you please provide more context and how we could measure said cache invalidation?

Oh, I believe you are talking recompiling files because of cycles. We have also made a lot of progress on this front but I think you might be misunderstanding the impact of cycles/file dependencies in Elixir.

In Elixir, we have compile-time and runtime dependencies. But in Gleam, because of imports and type inference (I assume), every Gleam dependency behaves as if it was a compile-time dependency in Elixir. I have updated the repository with two new repos to show exactly what I mean: https://github.com/josevalim/langcompilebench

Both repos have 100 modules, which go from A1->A2->...->A100 (Elixir) to a1->a2->...a100 (Gleam). When we change a100.ex in Elixir, it recompiles one file. When we change a100.gleam, it recompiles 100 files. So while you don't have cycles (which is definitely a plus one for Gleam, I'd personally mention it as a benefit), every dependency between modules in Gleam clauses recompilation, but in Elixir, only compile time ones do.

And even though Elixir can have cycles, the cycle only affects the compile-time dependencies in said cycle (note you cannot have a cycle of only compile-time deps). For example, if you clone https://github.com/livebook-dev/livebook and run `mix xref graph --format cycles`, you will find a 150 long cycle, but changing a file in this cycle leads only 3 files to recompile, because those are the compile-time dependencies.

On the other hand, in Gleam, even without cycles, if I change file `a`, it will force everyone that imports `a` to recompile, and so on recursively. The recursive behaviour happens only in Elixir for compile-time dependencies, and it stops as soon as it sees a runtime dependency, but in Gleam "compile-time" is the only dependency type.

If anything is wrong in my understanding, please let me know, but I wouldn't be surprised if Gleam currently triggers more recompilations than Elixir. If you run `mix xref graph --format stats` in Livebook, you will find only 17% of the dependencies between files trigger direct recompilation (vs 100% in Gleam).

Ideally we would get a large-ish project in both languages, change each file in it, and recompute how many compilations it triggers on average, but that may be hard to come by. Meanwhile I'd gladly accept PRs to the `langcompilebench` repository that shows a case I may be missing. I am only asking for us make these decisions on data and examples we can verify, even if somewhat artificial (real-world is better).

> Elixir generates more code and will compile multiple times to resolve macros, so the cost of the backend is more impactful.

Just to clarify, because the wording is ambiguous, Elixir will compile one additional time to perform macro expansion, regardless of how many macros. It is twice altogether. It is also worth saying that in Elixir, both `def` and `defmodule` are macros, so even if your claim is that the initial repository is not representative of Elixir code (which I agree), each file in that repo is calling 101 macros, and the times between Gleam and Elixir were still comparable.

josevalim: I have spent the last few days playing with a web app in Gleam to better understand when recompilations happen in Gleam. For example, if you are using [squirrel](https://hexdocs.pm/squirrel/) (which is very neat btw), every time you add/remove/change a query, all callers of that given `sql` module will be recompiled, regardless of the query. So in a codebase that uses models (the ones that invoke the queries) and controllers (the ones that invoke the models), then all controllers which calls models in that group of queries are recompiled (I did this by looking at the .beam timestamps on disk, so if I missed something, please let me know!).

---

To try to better visualize this, let's consider a small tree. I am using a balanced tree, cause it is easier to reason about. Imagine you have this file structure, A1/A2/A3 depends on A which depends on R and so on:

```mermaid
graph BT
    A1 --> A
    A2 --> A
    A3 --> A
    B1 --> B
    B2 --> B
    B3 --> B
    C1 --> C
    C2 --> C
    C3 --> C
    A --> R
    B --> R
    C --> R
```

In Gleam, if you change R (which would be like `myapp/sql`), all nodes downstream are compiled. If you change A, A1/A2/A3 are recompiled. And so on. So if you change a file at random, the average amount of files recompiled per change is `(13 + 3*4 + 9*1) / 13`, which is 2.61.

However, in Elixir, it won't propagate unless you have compile-time dependencies. So even though Ecto uses macros, changing a query does not lead to compile-time propagation. So the default average is 1. But apps will certainly have compile-time dependencies too, so let's imagine `A`, `B`, and `C` depend on `R` as a compile-time dependency (like Phoenix apps all have `use MyAppWeb, :controller`) and mark it in red. We end-up with this:

```mermaid
graph BT
    A1 --> A
    A2 --> A
    A3 --> A
    B1 --> B
    B2 --> B
    B3 --> B
    C1 --> C
    C2 --> C
    C3 --> C
    A -->|compile| R
    B -->|compile| R
    C -->|compile| R

    linkStyle 9,10,11 stroke:red
```

This makes it so 3 out of 12 edges are compile-time edges (which is a rate of 25%, which is higher than Livebook's 17% average). Now, when R changes, it compiles A, B and C, but that's it. So our recompilations per file average becomes `(4 + 3*1 + 9*1) / 13`, which is 1.23. Less than half of Gleam's.

But what happens if we introduce a cycle? Let's say that R depends on C2:

```mermaid
graph BT
    A1 --> A
    A2 --> A
    A3 --> A
    B1 --> B
    B2 --> B
    B3 --> B
    C1 --> C
    C2 --> C
    C3 --> C
    A -->|compile| R
    B -->|compile| R
    C -->|compile| R
    R --> C2
    linkStyle 9,10,11 stroke:red
```

What this means is that, anything in that path (C and C2) will trigger R, so our average is:

* R changes, we compile: R, A, B, C
* C changes, we compile: A, B, C
* C2 changes, we compile: A, B, C, C2

Everything else stays the same, so we have `(4 + 3 + 4 + 10) / 13`, which is 1.61 and still well below Gleam's average for this tree.

> Note: when R depends on C2, what we call it a compile-connected dependency, and we have tooling in `mix` to help find them!

Furthermore, even if we make A2 and B2 cycles, similar to C2, the average is 2.4 and still below Gleam's. And that's already a very unlikely scenario: remember that Livebook has 300 files and a single cycle with 2 compile-time dependencies, so it is very unexpected for a tree with 13 files to have 3 cycles with 3 compile-time dependencies each.

Once again, I am not saying this is representative (it may be, it may not be), but I hope this illustrates the situation is quite more nuanced than claimed, and that from an analytical point of view, it does not necessarily hold that cache invalidation is a bigger issue in Elixir than in Gleam, even in the presence of cycles.

---

Furthermore, I have looked deeper into Elixir's compilation and that fact we compiled modules definitions, so [I added an interpreted option](https://github.com/elixir-lang/elixir/pull/15087). I could not see meaningful changes in the benchmarks above or in Livebook, but Remote's codebase, which has above 1 million lines of real-world code, compiles 5x faster than v1.19 and 3x faster than main. Initial analysis shows the issue is not the backend, but code loading, but the new option should now address any theoretical (and practical!) concerns related to that.

I will no longer press on this issue. I believe I have provided all I could from my side: benchmarks, data, pull requests, and explanations.

You are welcome to close this pull request or edit it as you seem fit. :) I believe it is fair to keep this line until Elixir v1.20 is out, but if you plan to keep it for longer, I hope we will have ways to validate the claims, so we can all collaborate on the topic. Thank you!

michallepicki: If there's no benchmarks provided, maybe it's worth rewording this sentence to something less absolute? E.g.
> We've found in our testing that Gleam compiles faster than Elixir.

I write Elixir day to day but I do love Gleam as well, both have their advantages and disadvantages, and I think both have a place in the BEAM ecosystem :) To expand on some of the already brought up points, and add a few more from my POV:

To me the feeling that Gleam compiles faster than Elixir is most apparent when modifying an existing codebase with an LSP in the editor. The Gleam LSP does not need to run the Erlang compiler for it to work, only part of the whole "pipeline" is needed to surface type errors etc. The Gleam LSP is super responsive and reliable, while with Elixir I am not using it (I'm either using Zed without LSP, or VSCode based editors with a [hacky hard to set up tree-sitter based highlighter extension](https://marketplace.visualstudio.com/items?itemName=AlecGhost.tree-sitter-vscode)).

There is a tradeoff of the convenience of having metaprogramming features (as opposed to having to explicitly run some codegen tasks in Gleam) vs the overhead it adds on recompilations due to "wasted" repeated work even when the result is the same. And in a real world codebase where compile dependencies between modules aren't always perfectly minimal, this not only makes recompilation slightly slower without apparent reason in a unnoticed-for-months way, it sometimes causes serious problems, there's a reason why there's been so many blog posts and conference talks about reducing compile dependencies and using `mix xref` to debug such issues, I think most Elixir projects hit such issues eventually. In Gleam the compiler will just yell at you? :) (but it's also limiting code organization and may be considered less convenient, btw see also https://github.com/gfngfn/Sesterl/issues/38 )

Similarly, Elixir has a convenient config system, but modifying the config will trigger whole project recompilation (understandably, hard to track what effect it could have without limiting the config functionality). Gleam? Most people probably use only runtime config with loading things like env variables or toml files at startup, no recompilation needed (but I do like the Elixir config a lot)

Maybe this one is silly but I think Gleam emits much less Erlang code than Elixir? Maybe I'm wrong but even a simple `variable.dot_access` in Elixir is a case with matching on the variable being a map, and try-rescue with an attempt to call a function in case the variable is a module. So it's likely that the time spent in the Erlang compiler part (sans the filesystem writing, BEAM startup, file reading and Erlang parsing overhead) is lower for Gleam. But also Gleam emits a lot of variable re-assignments, which sometimes hits slow paths in the Erlang compiler, I saw issues on the Erlang/OTP repo for some pathological cases (<3 to the OTP team for fixing them btw).

AFAIK there's still a lot of performance gains left on the table in the Gleam compiler. Most importantly I think the Gleam compiler doesn't make much use of parallel compilation yet? Some of the module graph recompilation optimizations @josevalim mentioned could also apply in Gleam. Not to mention any lower level things in Rust, like leveraging mutability more, organizing memory layouts for cache efficiency, reducing dynamic memory allocations with arenas... Probably more optimizations will come once there's need for them for working bigger real world codebases, but having a simpler compiler codebase is also worth preserving

The Elixir compiler is not slow overall, and it is true that a lot of good optimization work was put into it. Some upcoming typechecker features like exhaustiveness checking might probably slow it down a bit but we'll see by how much and how it turns out, I know there's no guarantees on some of the typechecker features being added at all, maybe it won't be feasible. Fingers crossed though, I enjoy the Gleam typechecker and would love more type based checks in Elixir as well

To summarize, I think Gleam compilation is probably often faster because it's a simpler language, and (also thanks to some limitations or not having some functionality) the compiler can be implemented in more efficient Rust. The Elixir compiler architecture is better for providing its metaprogramming features, but also for performance in some ways, and the Elixir compiler is doing some optimizations that Gleam doesn't do (yet?). The Gleam typechecker is based on a standard, proven to be fast enough HM type system that is making it clearly a statically typed language, while the Elixir typechecker is also an impessive gradual typing research project, bringing as much type safety as possible, but keeping the language ergonomics dynamic.

hayleigh-dot-dev: > Let's undo this change in this PR so we can merge the other changes until there's new benchmarks.

Let's be respectful of this and move the discussion to another venue so we can get the other changes merged and deployed 🙏.

josevalim: > To me the feeling that Gleam compiles faster than Elixir is most apparent when modifying an existing codebase with an LSP in the editor. The Gleam LSP does not need to run the Erlang compiler for it to work, only part of the whole "pipeline" is needed to surface type errors etc

Excellent! But that doesn't mean Gleam compiles faster than Elixir, it means Gleam can provide much (if not all) of its LSP features faster than Elixir because it requires only the front-end while Elixir has to either compile or interpret code. If the claim was that "Gleam LSP is faster than Elixir LSP", I would have a hard time disagreeing with that.

>  there's a reason why there's been so many blog posts and conference talks about reducing compile dependencies and using mix xref to debug such issues

I'd say this is partially missing the argument of the diagrams above. The reason why we can use `mix xref` in Elixir is because Elixir makes a distinction between runtime and compile-time dependencies. Gleam AFAIK doesn't, everything is compile-time, so Elixir has an opportunity to improve over Gleam's baseline.

Alternatively, you could imagine an Elixir version where every call between modules is a compile-time dependency. This means we wouldn't have cycles, `xref`'s would be gone and there'd be no need to talk about it, but now recompilations are way more frequent.

> To summarize, I think Gleam compilation is probably often faster because it's a simpler language, and (also thanks to some limitations or not having some functionality) the compiler can be implemented in more efficient Rust.

If that's the case, then we should be able to have benchmarks or some metric that shows it! And I did provide such benchmark initially, but it will be outdated by the next Elixir release.

My whole point is: let's move beyond what we *think* is happening, towards what we can actually measure. If Gleam can be parallelized, better utilize memory layouts, then let's have a shared place where we can see those improvements! It is clear we all have opinions on why we think one is faster than the other and I tried my best to bring some actual data into the discussion, otherwise it will only lead to endless speculation.

michallepicki: > > Let's undo this change in this PR so we can merge the other changes until there's new benchmarks.

> Let's be respectful of this and move the discussion to another venue so we can get the other changes merged and deployed 🙏.

But also it may feel disrespectful towards Elixir to keep this statement line without providing benchmarks to back it up

josevalim: Sorry @hayleigh-dot-dev, I didn't see your reply before my comment above. As said before:

> You are welcome to close this pull request or edit it as you seem fit. :)

I have enabled permission to edit the pull request and you folks are welcome to do as you prefer. I am unsubscribing from the issue, so you won't hear from me anymore, but I really hope there will be some initiative to bring visibility into this. Apologies for any misunderstanding.

PR #602, opened 2026-02-26
add wisp dev server

graphiteisaac:
Cheerio!

With the new watchexec package, running a wisp server and auto-rebuilding the site is a breeze, so I thought it was time to tap into the Gleam ecosystem for this, instead of using my currently very cursed local setup. Feels so nice to just run `gleam dev` and not think about it again.

I've gone to some level of thought (as far as my peanut brain can go, anyway) for static assets, so we don't have to rebuild the entire website every time a single line of CSS is changed, for example.


dev/website_dev.gleam
```diff
@@ -0,0 +1,136 @@
...
+}
+
+fn rebuild_full(_paths: List(String)) {
+  let _ = website.build_site()
+  wisp.log_info("Rebuilt full website")
+}
+
+fn rebuild_assets(changed_paths: List(String)) {
+  list.each(changed_paths, fn(path) {
+    let dest = string.replace(path, each: "/priv", with: "/dist")
```
lpil: This could do weird stuff if someone has a highler level `/priv*` directory. Can we strip the current working directory from the path instead? 🙏 

Something like this maybe

```gleam
fn rebuild_assets(cwd: String, changed_paths: List(String)) {
  list.each(changed_paths, fn(path) {
    let assert Ok(#("", shortened_path)) = string.split_once(path, cwd <> "/priv")
      as { path <> " not within priv dir" }
    // ...
```

lpil:
I merged this to main!

PR #630, opened 2026-04-10
Basic guides implementation

graphiteisaac:
Wahoo! It's time to get a better guides layout done. This uses the table-of-contents layout that was originally built for the guides page that was then bought ahead of schedule. 

This uses an inline SVG which I know Louis has not historically been keen on, but we're using it for a super cool reason, that being to have 3 colourations of the same icon without duplicating the actual SVG path markup.

I also added a read_directory function to the fs module to read guides from the respective directory, rather than hard-coding them in as we currently do. Feels appropriate to use djot (and frontmatter) for this system, to make it extra-approachable and easy to PR into.


src/website/guides.gleam
```diff
@@ -0,0 +1,163 @@
...
+
+pub fn target_string(target: Target) -> String {
+  case target {
+    Javascript -> "JavaScript"
+    Beam -> "BEAM"
+    Any -> "Any"
+  }
+}
+
+pub fn target_icon(target: Target, i: Int) {
```
lpil: The inline SVG swaps having 3 cacheable copies of it for having (0-2)*num_guides uncachable copies, that's much worse!

src/website/guides.gleam
```diff
@@ -0,0 +1,163 @@
...
+  )
+  use guide <- result.try(decode_frontmatter(path, parsed_frontmatter))
+  let content = jot.parse(content)
+  Ok(Guide(..guide, slug:, content:))
+}
+
+pub fn target_string(target: Target) -> String {
+  case target {
+    Javascript -> "JavaScript"
+    Beam -> "BEAM"
```
lpil: Lets call this Erlang, since the build tool and compiler do.

src/website/guides.gleam
```diff
@@ -0,0 +1,163 @@
...
+}
+
+fn toml_to_snag(result: Result(a, tom.GetError)) -> snag.Result(a) {
+  result.map_error(result, fn(e) { snag.new(string.inspect(e)) })
+}
+
+fn read(path: String) -> snag.Result(Guide) {
+  let slug = case string.ends_with(path, ".djot") {
+    // The length of ".djot", not my favourite way of doing this, but it works.
+    True -> string.drop_end(path, 5)
```
lpil: Can use this now: https://hexdocs.pm/gleam_stdlib/gleam/string.html#remove_suffix

✨

src/website/page.gleam
```diff
@@ -2697,6 +2751,34 @@ pub fn externals_guide(ctx: site.Context) -> snag.Result(fs.File) {
...
+
+  [
+    element.unsafe_raw_html(
+      "",
+      "article",
+      [class("prose")],
+      jot.document_to_html(guide.content),
+    ),
+  ]
+  |> table_of_contents_page_layout(table_of_contents, meta, ctx)
```
lpil: Most guides will be very short, just a paragraph or two, so the table of contents layout won't be a good fit.

Maybe we could set the layout with the frontmatter?

lpil (commented):
Amazing!!! I've left some tiny notes inline. 🙏

src/website/page.gleam
lpil: Oh wait maybe we should do it based upon whether there are headings! I can take this if you want.

lpil:
I'm confused, I pushed a commit but I can't see it here 🤔

lpil:
ah now it shows. GitHub bug I presume

lpil (approved):
Thanks Isaac!!!

PR #631, opened 2026-04-11
Minor fixes to gleam.toml documentation

folospior:
Hi there!

Amazing work on the gleam.toml documentation, just wanted to nitpick a little:

- **fix git dependency incorrect field name**
  - Git dependencies use the `git` and `ref` fields, not the `path` field
- **fix incorrect wording in dev dependencies description**
  - Dev dependencies were described as "used in tests" - that's not true, they're used in development. Tools like squirrel and lustre_dev_tools are also used as dev dependencies
- **fix invalid link in internal modules**
  - There were angle brackets around the link - these made the link unclickable

### DIFF
--- a/documentation/gleam-toml.djot
+++ b/documentation/gleam-toml.djot
@@ -101,7 +101,7 @@ attacks from malicious actors.
 gleeunit = ">= 1.0.0 and < 2.0.0"
 ```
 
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thanks!

PR #632, opened 2026-04-17
feat: change link to wall of sponsors on homepage

george-grec:
this should make it so people stay on the homepage and read a bit about the other core members they can support :)


lpil (approved):
Thank you!

PR #634, opened 2026-04-30
Document place for external tools' config in `gleam.toml`

ankddev:
* Closes #603.

I believe that explanation and example are good enough


giacomocavalieri (approved):

lpil (approved):
Thank you!!!

PR #636, opened 2026-05-11
Fix invalid Elixir block syntax in Gleam for Elixir users cheatsheet

### LINKED ISSUE #298: Invalid Elixir block syntax in "Gleam for Elixir Users"
I was reading the Blocks section of the [Gleam for Elixir Users](https://gleam.run/cheatsheets/gleam-for-elixir-users/#blocks) and I was surprised by this syntax — when I tried it in `iex`, it's not valid Elixir

```elixir
x = do
  print(1)
  2
end
```

([code link](https://github.com/gleam-lang/website/blob/main/cheatsheets/gleam-for-elixir-users.md?plain=1#L441-L444))

Screenshot 2024-02-07 at 10 27 19

Maybe there is a syntax, but I couldn't find it 🤔 

(I saw `1.14` in the screenshot, but I tried `1.15` and `1.16` too: same problem)

brone1323:
Fixes #298

## What

The Blocks section of the Gleam for Elixir users cheatsheet showed this code as valid Elixir:

```elixir
x = do
  print(1)
  2
end
```

This is **not valid Elixir**. `do/end` in Elixir is only valid as a keyword argument to macros such as `def`, `if`, and `case` — it cannot be used to group expressions in a bare variable assignment.

## Fix

- Updated the code example to use parentheses, which is the correct Elixir syntax for grouping expressions in an assignment context
- Updated the description text to accurately explain when `do/end` vs parentheses are used in Elixir

The corrected example:

```elixir
x = (
  IO.puts(1)
  2
)
```

This matches what `inoas` verified in the issue comments (referencing the [Elixir syntax reference](https://hexdocs.pm/elixir/1.16.1/syntax-reference.html#blocks)).

### DIFF
--- a/src/website/cheatsheet.gleam
+++ b/src/website/cheatsheet.gleam
@@ -919,21 +919,31 @@ fn main() {
     html.h2([attr.id("blocks")], [html.text("Blocks")]),
     html.h4([attr.id("elixir-12")], [html.text("Elixir")]),
     html.p([], [
-      html.text("In Elixir expressions can be grouped using "),
+      html.text(
+        "In Elixir expressions can be grouped using parentheses. The ",
+      ),
       html.code([], [html.text("do")]),
-      html.text(" and "),
+      html.text("/"),
       html.code([], [html.text("end")]),
-      html.text("."),
+      html.text(
+        " syntax is only valid as a keyword argument to macros such as ",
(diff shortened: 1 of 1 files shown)


graphiteisaac:
Thanks so much!

PR #641, opened 2026-06-03
Stylize macOS with proper capitalization

louwers:
macOS is stylized with a lowercase m since 2016. https://en.wikipedia.org/wiki/MacOS

Built locally to verify it works.


giacomocavalieri (approved):

lpil (approved):
Thanks!

PR #645, opened 2026-06-04
Update FAQ with Elixir and Gleam differences

josevalim:
Thank you for the recent updates, I have ported some additional notes we had previously discussed in #595.

### DIFF
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -150,12 +150,14 @@ popular and a great language!
 Here’s a non-exhaustive list of differences:
 
 - Elixir is gradually typed, while Gleam is fully statically typed.
-- Elixir's type system does not have generics, while Gleam's type system does.
+- Elixir's type system does not support user provided type annotations at the moment,
+  while Gleam's type system does.
 - Elixir has a powerful macro system, Gleam has no metaprogramming features.
 - Elixir’s compiler is written in Erlang and Elixir, Gleam’s is written in Rust.
 - Gleam has a more traditional C family style syntax.
 - Elixir has a namespace for module functions and another for variables, Gleam
-  has one unified namespace (so there’s no special `fun.()` syntax).
+  has one unified namespace (so there’s no special `fun.()` syntax, but in Gleam
+  variable names can shadow function names).
 - Gleam standard library is distributed as Hex packages, which makes
   interoperability with other BEAM languages easier.
 - Elixir is a larger language, featuring numerous language features not present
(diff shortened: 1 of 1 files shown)


documentation/frequently-asked-questions.djot
```diff
@@ -150,12 +150,14 @@ popular and a great language!
 Here’s a non-exhaustive list of differences:
 
 - Elixir is gradually typed, while Gleam is fully statically typed.
-- Elixir's type system does not have generics, while Gleam's type system does.
+- Elixir's type system does not support user provided type annotations at the moment,
```
josevalim: Elixir's limitation here is actually deeper than generics.

lpil: Thank you. Does it support generics now then and I misunderstood?

lpil (commented):
Left a question inline

documentation/frequently-asked-questions.djot
josevalim: It doesn't support generics (they do in the theory but we haven't implemented it internally) but they would have limited use in our type system anyway unless we introduce type signatures. So to me it goes like this:

1. Elixir has no types -> we add gradual typing
2. Elixir has no type signatures -> we add type signatures
3. Elixir has no generics -> we add generics

You can totally leave it as is but in my opinion not having type signatures is a bigger limitation of Elixir when compared to Gleam.

lpil: > they would have limited use in our type system anyway unless we introduce type signatures

That may be true of Elixir specifically, but that's not how generics work more widely, so folks unfamiliar with Elixir would not be able to infer from the lack of annotations that generics are not supported. Gleam and other ML languages don't require any annotations for generics, for examples.

I'll expand on this entry to make it clearer.

josevalim: That reframes the discussion though. It would be valid to say that "the Elixir type system does not infer generics" but on the other hand, we can infer non-discriminated unions, intersections, and negations, which the Gleam type system cannot. I am not sure it is a useful distinction to call out.

josevalim: @lpil I have reverted this change because I am truly fine with shipping the current description. It is correct, I just don't find it a useful distinction, but fair game if you believe it is! Thanks for the discussion! ❤️

documentation/frequently-asked-questions.djot
```diff
@@ -150,12 +150,14 @@ popular and a great language!
...
+  while Gleam's type system does.
```
```suggestion
- Elixir's type system does not have generics, while Gleam's type system does.
```

lpil (approved):
Thanks José!!

PR #646, opened 2026-06-04
Refine comparison of Elixir and Gleam type systems

josevalim:
Remove duplication, make clear what is a language choice vs a type system choice, and remove unclear sentences.

### DIFF
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -163,23 +163,12 @@ Here’s a non-exhaustive list of differences:
     annotations, but they are not required.
   - Elixir does not support generics, while Gleam does. This makes Gleam's type
     system more precise in some respects.
-  - Elixir's type system does not support recursive types, while Gleam does.
+  - Elixir does not support recursive types, while Gleam does.
     This makes some data structures impossible to represent in Elixir's type
     system.
-  - Gleam's type system errs on the side of false negatives, whereas Elixir's
-    type system chooses to prefer false positives. This means Gleam may reject
-    a program that is actually safe, but Elixir may accept a program that
-    crashes at runtime due to a type error.
-  - Elixir's type system does not integrate with any other system. Gleam's type
+  - Elixir does not currently integrate with any other system. Gleam's type
     system generates BEAM typespecs, TypeScript definitions, and API type
     information can be exported to be used by other external tools.
-  - Elixir's type system is the product of novel research specifically for
-    Elixir. Gleam's type system is based on the same Hindley-Milner type system
-    that underpins OCaml, Haskell, F#, etc. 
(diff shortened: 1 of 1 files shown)


lpil:
Hi José! Why have you removed these points? Do you think they are not accurate?

documentation/frequently-asked-questions.djot
```diff
@@ -163,23 +163,12 @@ Here’s a non-exhaustive list of differences:
     annotations, but they are not required.
   - Elixir does not support generics, while Gleam does. This makes Gleam's type
     system more precise in some respects.
-  - Elixir's type system does not support recursive types, while Gleam does.
+  - Elixir does not support recursive types, while Gleam does.
     This makes some data structures impossible to represent in Elixir's type
     system.
-  - Gleam's type system errs on the side of false negatives, whereas Elixir's
```
josevalim, line 169: This is the only actual removal, cause I don’t believe it is accurate. The Elixir type system is more expressive and there are papers that prove set theoretic types are a superset of HM (intersections are enough for that), which should result in fewer false positives and false negatives.

Furthermore our current implementation minimizes false positives by design, so it is inaccurate to say err on that side.

There is an argument about dynamic but that has already been specified above as it being gradually typed.

lpil: AH yes, it does make it sound like the intent of the design is to favour false negatives, but that is not the case.

lpil (approved):
Thank you!

documentation/frequently-asked-questions.djot
```diff
@@ -163,23 +163,21 @@ Here’s a non-exhaustive list of differences:
...
-    type system chooses to prefer false positives. This means Gleam may reject
-    a program that is actually safe, but Elixir may accept a program that
-    crashes at runtime due to a type error.
+  - Elixir's type checker is more likely to have false-negatives, while Gleam
+    is more likely to have false-positives. This means Gleam may reject a
+    program that is actually safe, but Elixir may accept a program that crashes
+    at runtime due to a type error.
```
josevalim, line 173: “This means Gleam may reject a program that is actually safe, but Elixir may accept a program that crashes at runtime due to a type error.”

If there is a type error, then we will emit a type violation. The scenarios we crash is because we have not emitted typing violations (same as Gleam, they are both sound). Dynamic is not a source of unsoundness. The errors happen because the type system does not capture a particular idiom with precision, so it is not part of the type.

I think we need to describe precisely what we mean about false positives and false negatives, because we may be talking about different things. I will send an amendment later.

josevalim:
Btw, what do you mean by: “ This means Elixir's types can be more precise in some respects, but it cannot distinguish between types that are semantically distinct but structurally the same”? I’d love to learn more.

PR #647, opened 2026-06-04
Update FAQ on Elixir and Gleam type systems

josevalim:
I am trying to organize the differences so they work more top to bottom. We first start with the overall description of the systems, and then we break it down.

I am also careful about the words here, trying to distinguish between the type system and the implementation, as I can see this becoming a more general discussion between set-theoretic types and HM, so I want to make it clear what is Elixir's fault and what is a type system limitation.

A couple notes:

* Elixir is type-theoretic too, so I removed that altogether
* Only certain aspects of Elixir's type system are novel. Overall set theoretic types are a few decades old :)
* Set-theoretic types are more expressive than HM and it is proven intersection types are a superset of HM inference (I can dig the exact paper later). However, we cannot affirm that for Elixir in particular because it is still in development

### DIFF
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -164,8 +164,6 @@ Here’s a non-exhaustive list of differences:
     annotations, but they are not required.
   - Elixir does not support generics, while Gleam does. This makes Gleam's type
     system more precise in some respects.
-  - Elixir does not support recursive type declarations, while Gleam does.
-    This makes some data structures impossible to represent in Elixir.
   - Elixir does not support nominal types and therefore cannot distinguish between
     types that are semantically distinct but structurally the same, while Gleam can.
   - Elixir does not integrate with any other type system. Gleam generates
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -167,10 +167,10 @@ Here’s a non-exhaustive list of differences:
     annotations, but they are not required.
   - Elixir does not support generics, while Gleam does. This makes Gleam's type
     system more precise in some respects.
-  - Elixir does not support recursive types, while Gleam's type system does.
+  - Elixir does not support recursive types, while Gleam does.
     This makes some data structures impossible to represent in Elixir.
-  - Elixir does not integrate with any other system. Gleam's type
-    system generates BEAM typespecs, TypeScript definitions, and API type
+  - Elixir does not integrate with any other type system. Gleam generates
+    BEAM typespecs, TypeScript definitions, and API type
     information can be exported to be used by other external tools.
   - Elixir's type inference aims to minimize false positive warnings, emitting
     violations only when a program always fails at runtime. Gleam's type inference
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -152,23 +152,22 @@ Here’s a non-exhaustive list of differences:
 - Elixir and Gleam both have compile type type checking, but their respective
   type systems are very different, so they offer a different development
   experience.
-  - Elixir has a set-theoretic type system, where types can be composed with
-    unions, intersections, and negations. Gleam's type system is based on
+  - Elixir has a set-theoretic type system. Gleam's type system is based on
     the same Hindley-Milner type system that underpins OCaml, Haskell, F#, etc.
   - Elixir's type system is gradual, so portions of the codebase can be
     dynamically typed. Gleam is always fully statically typed.
   - Elixir's type system is structural, while Gleam's type system is nominal.
-    Elixir's types can be more precise in many respects. On the other hand,
-    Elixir's types are in active development and lacks common features (mentioned below).
-  - Elixir does not have opaque types and therefore cannot distinguish between
-    types that are semantically distinct but structurally the same, while Gleam can.
+    Elixir's types can be more precise in many respects.
+  - Elixir's type system is in development, while Gleam's is more mature.
   - Elixir does not support type annotations, so the programmer cannot restrict
     the type beyond what Elixir infers them to be. Gleam supports type
     annotations, but they are not required.
   - Elixir does not support generics, while Gleam does. This makes Gleam's type
     system more precise in some respects.
-  - Elixir does not support recursive types, while Gleam does.
+  - Elixir does not support recursive type declarations, while Gleam does.
     This makes some data structures impossible to represent in Elixir.
+  - Elixir does not support nominal types and therefore cannot distinguish between
+    types that are semantically distinct but structurally the same, while Gleam can.
   - Elixir does not integrate with any other type system. Gleam generates
     BEAM typespecs, TypeScript definitions, and API type
     information can be exported to be used by other external tools.
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -159,7 +159,7 @@ Here’s a non-exhaustive list of differences:
     dynamically typed. Gleam is always fully statically typed.
   - Elixir's type system is structural, while Gleam's type system is nominal.
     Elixir's types can be more precise in many respects. On the other hand,
-    Elixir is still in active development and lacks common features (discussed below).
+    Elixir's types are in active development and lacks common features (mentioned below).
   - Elixir does not have opaque types and therefore cannot distinguish between
     types that are semantically distinct but structurally the same, while Gleam can.
   - Elixir does not support type annotations, so the programmer cannot restrict
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -152,6 +152,7 @@ Here’s a non-exhaustive list of differences:
 - Elixir and Gleam both have compile type type checking, but their respective
   type systems are very different, so they offer a different development
   experience.
+
   - Elixir has a set-theoretic type system. Gleam's type system is based on
     the same Hindley-Milner type system that underpins OCaml, Haskell, F#, etc.
   - Elixir's type system is gradual, so portions of the codebase can be
--- a/documentation/frequently-asked-questions.djot
+++ b/documentation/frequently-asked-questions.djot
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
 - Elixir and Gleam both have compile type type checking, but their respective
   type systems are very different, so they offer a different development
   experience.
+  - Elixir has a set-theoretic type system, where types can be composed with
+    unions, intersections, and negations. Gleam's type system is based on
+    the same Hindley-Milner type system that underpins OCaml, Haskell, F#, etc.
   - Elixir's type system is gradual, so portions of the codebase can be
     dynamically typed. Gleam is always fully statically typed.
   - Elixir's type system is structural, while Gleam's type system is nominal.
-    This means Elixir's types can be more precise in some respects, but it
-    cannot distinguish between types that are semantically distinct but
-    structurally the same.
+    Elixir's types can be more precise in many respects. On the other hand,
+    Elixir is still in active development and lacks common features (discussed below).
+  - Elixir does not have opaque types and therefore cannot distinguish between
+    types that are semantically distinct but structurally the same, while Gleam can.
   - Elixir does not support type annotations, so the programmer cannot restrict
     the type beyond what Elixir infers them to be. Gleam supports type
     annotations, but they are not required.
   - Elixir does not support generics, while Gleam does. This makes Gleam's type
     system more precise in some respects.
-  - Elixir's type system does not support recursive types, while Gleam's type
-    system does.
-    This makes some data structures impossible to represent in Elixir's type
-    system.
-  - Elixir's type checker is more likely to have false-negatives, while Gleam
-    is more likely to have false-positives. This means Gleam may reject a
-    program that is actually safe, but Elixir may accept a program that crashes
-    at runtime due to a type error.
-  - Elixir's type system does not integrate with any other system. Gleam's type
+  - Elixir does not support recursive types, while Gleam's type system does.
+    This makes some data structures impossible to represent in Elixir.
+  - Elixir does not integrate with any other system. Gleam's type
     system generates BEAM typespecs, TypeScript definitions, and API type
     information can be exported to be used by other external tools.
-  - Elixir's type system is the product of novel research specifically for
-    Elixir. Gleam's type system is based on the same Hindley-Milner type system
-    that underpins OCaml, Haskell, F#, etc. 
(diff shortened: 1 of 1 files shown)


documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
+    unions, intersections, and negations. Gleam's type system is based on
+    the same Hindley-Milner type system that underpins OCaml, Haskell, F#, etc.
   - Elixir's type system is gradual, so portions of the codebase can be
     dynamically typed. Gleam is always fully statically typed.
   - Elixir's type system is structural, while Gleam's type system is nominal.
-    This means Elixir's types can be more precise in some respects, but it
-    cannot distinguish between types that are semantically distinct but
-    structurally the same.
+    Elixir's types can be more precise in many respects. On the other hand,
+    Elixir is still in active development and lacks common features (discussed below).
```
```suggestion
    Elixir's types are in active development and lacks common features (mentioned below).
```

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
+  - Elixir does not have opaque types and therefore cannot distinguish between
+    types that are semantically distinct but structurally the same, while Gleam can.
```
josevalim: This is an improvement over the previous one which read as a general limitation of structural vs nominal. If I am understanding what this is implying, it can be done, just not currently implemented.

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
+  - Elixir does not integrate with any other system. Gleam's type
     system generates BEAM typespecs, TypeScript definitions, and API type
     information can be exported to be used by other external tools.
-  - Elixir's type system is the product of novel research specifically for
-    Elixir. Gleam's type system is based on the same Hindley-Milner type system
-    that underpins OCaml, Haskell, F#, etc. 
-  - Elixir's system is set-theoretic. Gleam's system is type-theoretic.
+  - Elixir's type inference aims to minimize false positive warnings, emitting
+    violations only when a program always fails at runtime. Gleam's type inference
+    will reject valid programs in favor of finding more bugs.
```
josevalim, line 175: I did my best to reduce the amount of unqualified "false positive" vs "false negative". Instead saying valid programs or false positive warnings.

Also because set theoretic types can be quite more expressive, I focused in particular in Elixir's type inference. We can revisit once we add type signatures.

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
-    system.
-  - Elixir's type checker is more likely to have false-negatives, while Gleam
-    is more likely to have false-positives. This means Gleam may reject a
-    program that is actually safe, but Elixir may accept a program that crashes
-    at runtime due to a type error.
-  - Elixir's type system does not integrate with any other system. Gleam's type
+  - Elixir does not support recursive types, while Gleam's type system does.
+    This makes some data structures impossible to represent in Elixir.
+  - Elixir does not integrate with any other system. Gleam's type
     system generates BEAM typespecs, TypeScript definitions, and API type
```
```suggestion
  - Elixir does not support recursive types, while Gleam does.
    This makes some data structures impossible to represent in Elixir.
  - Elixir does not integrate with any other type system. Gleam generates
    BEAM typespecs, TypeScript definitions, and API type
```

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
-    is more likely to have false-positives. This means Gleam may reject a
-    program that is actually safe, but Elixir may accept a program that crashes
-    at runtime due to a type error.
-  - Elixir's type system does not integrate with any other system. Gleam's type
-    system generates BEAM typespecs, TypeScript definitions, and API type
+  - Elixir does not support recursive types, while Gleam does.
+    This makes some data structures impossible to represent in Elixir.
+  - Elixir does not integrate with any other type system. Gleam generates
+    BEAM typespecs, TypeScript definitions, and API type
     information can be exported to be used by other external tools.
```
josevalim, line 172: This is a really good call out and I am glad you added it.

documentation/frequently-asked-questions.djot
josevalim: Also please double check the usage of the word "valid", as it can be quite loaded in typing discussions.

documentation/frequently-asked-questions.djot
lpil: Elixir does support recursive types, but its static type system does not.

documentation/frequently-asked-questions.djot
lpil: Thank you!

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
 - Elixir and Gleam both have compile type type checking, but their respective
   type systems are very different, so they offer a different development
   experience.
+  - Elixir has a set-theoretic type system, where types can be composed with
+    unions, intersections, and negations. Gleam's type system is based on
+    the same Hindley-Milner type system that underpins OCaml, Haskell, F#, etc.
```
lpil, line 157: How come you've added some detail about set theory but not about type theory? Not sure what the goal is with this change.

If you'd like to make it more like the Gleam sentence we could change it to mention that TypeScript and Python are existing set-theoretic languages.

documentation/frequently-asked-questions.djot
lpil: The change to introduce opaque types makes this statement incorrect as that feature is not why Gleam has this property while Elixir does not. Opaque types are also a form of nominal typing.

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
   - Elixir's type system is gradual, so portions of the codebase can be
     dynamically typed. Gleam is always fully statically typed.
   - Elixir's type system is structural, while Gleam's type system is nominal.
-    This means Elixir's types can be more precise in some respects, but it
-    cannot distinguish between types that are semantically distinct but
-    structurally the same.
+    Elixir's types can be more precise in many respects. On the other hand,
+    Elixir's types are in active development and lacks common features (mentioned below).
```
lpil: This is implying that Elixir will get these features in future, but that may not be the case, and I believe some of them are not planned at all.

When any additions are made to Elixir's type system then this documentation will be updated to reflect the current state of play.

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
-    This makes some data structures impossible to represent in Elixir's type
-    system.
-  - Elixir's type checker is more likely to have false-negatives, while Gleam
-    is more likely to have false-positives. This means Gleam may reject a
-    program that is actually safe, but Elixir may accept a program that crashes
-    at runtime due to a type error.
-  - Elixir's type system does not integrate with any other system. Gleam's type
-    system generates BEAM typespecs, TypeScript definitions, and API type
+  - Elixir does not support recursive types, while Gleam does.
+    This makes some data structures impossible to represent in Elixir.
```
```suggestion
  - Elixir does not support recursive type declarations, while Gleam does.
    This makes some data structures impossible to represent in Elixir.
```

Good call @lpil. WDYT of the above?

documentation/frequently-asked-questions.djot
lpil: Good point RE valid. I'll expand this to be more descriptive.

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
   - Elixir's type system is gradual, so portions of the codebase can be
     dynamically typed. Gleam is always fully statically typed.
   - Elixir's type system is structural, while Gleam's type system is nominal.
-    This means Elixir's types can be more precise in some respects, but it
-    cannot distinguish between types that are semantically distinct but
-    structurally the same.
+    Elixir's types can be more precise in many respects. On the other hand,
+    Elixir's types are in active development and lacks common features (mentioned below).
+  - Elixir does not have opaque types and therefore cannot distinguish between
+    types that are semantically distinct but structurally the same, while Gleam can.
```
josevalim: The current version is implying Gleam has opaque types which may not be the best phrasing. Perhaps you prefer:

```suggestion
  - Elixir does not have opaque type declarations and therefore cannot distinguish 
    between types that are semantically distinct but structurally the same,
    while Gleam can distinguish them through nominal typing.
```

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,29 @@ Here’s a non-exhaustive list of differences:
...
-    system does.
-    This makes some data structures impossible to represent in Elixir's type
-    system.
-  - Elixir's type checker is more likely to have false-negatives, while Gleam
-    is more likely to have false-positives. This means Gleam may reject a
-    program that is actually safe, but Elixir may accept a program that crashes
-    at runtime due to a type error.
-  - Elixir's type system does not integrate with any other system. Gleam's type
-    system generates BEAM typespecs, TypeScript definitions, and API type
+  - Elixir does not support recursive types, while Gleam does.
```
lpil: Elixir does support recursive types, but its static type system does not.

lpil (commented):
Please don't re-order the list, it is making it harder to review!

documentation/frequently-asked-questions.djot
josevalim: Good point. I will remove it. I want to avoid Python or TypeScript comparisons because people may over reach from them (TS is not sound, none of them are semantic set theoretic types).

josevalim (commented):
Thank you @lpil! I have accepted all of your feedback. I decided to avoid using the word opaque and mentioned nominal types exclusively.

documentation/frequently-asked-questions.djot
josevalim: Fixed. Instead I am saying that Elixir's type system is in development and Gleam's more mature. But I don't mind if you prefer to remove such sentence.

documentation/frequently-asked-questions.djot
josevalim: Done!

josevalim:
And for completeness, here are the papers I mentioned:

> [D. Leivant, "Polymorphic Type Inference," POPL 1983, pp. 88–98.](https://dl.acm.org/doi/abs/10.1145/567067.567077)

This paper introduced the notion of rank for polymorphic types and showed that Hindley–Milner polymorphism corresponds to rank-2 (quantifiers only at the left of arrows). It is the foundational reference for rank-stratified comparisons.

> [H. Barendregt, M. Coppo, M. Dezani-Ciancaglini, "A Filter Lambda Model and the Completeness of Type Assignment," Journal of Symbolic Logic 48(4), 1983, pp. 931–940](https://www.jstor.org/stable/2273659)

The foundational completeness result for the BCD intersection type system. The filter model construction here is the semantic basis for understanding how intersection types capture all typable behaviors, including those expressible with universal quantification.

---

Of course those papers compare with polymorphic type inference but we don't have generics in Elixir yet, so the results do not apply to Elixir-Gleam today. Once (if?) generics are added, those are proofs that rank-2 intersection type systems (which Elixir would be) are strictly more powerful than ML-style type systems (in the sense that they can type strictly more terms).

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,28 @@ Here’s a non-exhaustive list of differences:
 - Elixir and Gleam both have compile type type checking, but their respective
   type systems are very different, so they offer a different development
   experience.
+  - Elixir has a set-theoretic type system. Gleam's type system is based on
```
ruilopesm, line 156: It seems, based on Djot's [README](https://github.com/jgm/djot/blob/main/README.md) that a sublist must always be preceded by a blank line.

We might need to add one; otherwise, the formatting will be incorrect. The current deployment, from a previous related PR, is already showing us that :P

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,28 @@ Here’s a non-exhaustive list of differences:
...
-    This makes some data structures impossible to represent in Elixir's type
-    system.
-  - Elixir's type checker is more likely to have false-negatives, while Gleam
-    is more likely to have false-positives. This means Gleam may reject a
-    program that is actually safe, but Elixir may accept a program that crashes
-    at runtime due to a type error.
-  - Elixir's type system does not integrate with any other system. Gleam's type
-    system generates BEAM typespecs, TypeScript definitions, and API type
+  - Elixir does not support recursive type declarations, while Gleam does.
+    This makes some data structures impossible to represent in Elixir.
```
```suggestion
```

documentation/frequently-asked-questions.djot
```diff
@@ -152,32 +152,26 @@ Here’s a non-exhaustive list of differences:
...
+  - Elixir's type system is in development, while Gleam's is more mature.
   - Elixir does not support type annotations, so the programmer cannot restrict
     the type beyond what Elixir infers them to be. Gleam supports type
     annotations, but they are not required.
   - Elixir does not support generics, while Gleam does. This makes Gleam's type
     system more precise in some respects.
-  - Elixir's type system does not support recursive types, while Gleam's type
-    system does.
-    This makes some data structures impossible to represent in Elixir's type
-    system.
```
josevalim, line 169: I have removed this one because it is either incorrect or incomplete:

> This makes some data structures impossible to represent in Elixir's type system.

Elixir can represent them just fine, the type system always sees the expanded form. What Elixir cannot do today is to infer a recursive signature, but that's entering into type variables/universal quantification which we do mention elsewhere.

Recursive types play a role once we have type annotations which we already cover that we don't have.

documentation/frequently-asked-questions.djot
josevalim:
```suggestion

  - Elixir has a set-theoretic type system. Gleam's type system is based on
```

josevalim: Done, thank you!!

josevalim:
@lpil anything I can do to get this one merged? I have addressed all other comments. Thanks!

lpil:
Hi José! Sorry but I'm very busy with a lot of notifications, and I have spent an unusual large amount of time on these website wording PRs, so I've stopped pulling it to the front of the queue so it doesn't interfere with higher priority works that has been waiting longer. Hopefully I will get to this soon, thank you for understanding!

lpil:
I've greatly cut down this section to avoid any future problems, and so we don't need to update it further as Gleam and Elixir each improve over time.

PR #648, opened 2026-06-06
Fix typo in FAQ regarding Elixir and Gleam compile time type checking

charlieaten:
Fixes a typo in the Elixir vs. Gleam comparison docs, changing “compile type type checking” to “compile time type checking” for clarity and correctness.


lpil (approved):
Thank you!

PR #651, opened 2026-06-15
Correct sans io return type

ollie-dot-earth:
The example for the sans-io pattern had `Result(Nil, ApiError)` where it should be `Result(User, ApiError)`

also ran `gleam format`


ollie-dot-earth:
github is odd

ollie-dot-earth:
it wont let me reopen it so ig ill make a new one ._.

PR #652, opened 2026-06-15
Correct the return type in sans-io pattern example

ollie-dot-earth:
The example for the sans-io pattern had `Result(Nil, ApiError)` where it should be `Result(User, ApiError)`


lpil (approved):
Thank you!!

PR #656, opened 2026-06-23
Reconcile CPU architectures in precompiled binary installation examples

ngscheurich:
The instructions on the _Precompiled executable from GitHub_ pages (such as https://gleam.run/install/macos/gleam/github/) provide examples for downloading, verifying, and extracting the Gleam binary such that it is executable and ready to roll. However, the CPU architectures used throughout the examples are not consistent. Compare, for example, the download block versus the extract block:

> `wget https://github.com/gleam-lang/gleam/releases/download/v1.14.0/gleam-v1.14.0-aarch64-unknown-linux-musl.tar.gzk`

> `tar xf gleam-v1.14.0-x86_64-unknown-linux-musl.tar.gz`

A minor detail overall, I think, but also a small improvement that might prevent a newcomer from getting hung up!

I chose aarch64 here mostly because that's what the examples start with.

### DIFF
--- a/src/website/install.gleam
+++ b/src/website/install.gleam
@@ -485,13 +485,13 @@ Verify the checksum is correct. If this command shows a warning then delete
 both files and start again.
 
 ```
-sha256sum -c gleam-v1.14.0-x86_64-unknown-linux-musl.tar.gz.sha256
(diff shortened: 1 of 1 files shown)


lpil (approved):
Oops, thank you!

PR #657, opened 2026-06-24
Update escript instructions & package versions in example

ginkogruen:
I've updated the instructions for sharing your program via an escript file to use the new `gleam export escript` command.

While at that I've also updated the version numbers for the packages in the example.

I wouldn't normally have done this but I've thought since `gleam_stdlib` is now `1.0.0` it would be nice to have that reflected in there.


lpil (approved):
Fantastic, thank you very much!!

PR #658, opened 2026-06-24
Add install instructions for Nix and NixOS

0xda157:
Fixes #613 Fixes #612


lpil:
Excellent, thank you! I've amended the commit to optimise the SVG.

The new guides include several methods for installing with Nix. Would it make more sense to have these?

- Nix shell
- Nix home manager
- NixOS configuration

There seems to be 2 commands for Nix shell, and it looks like you're supposed to run both. Is that right?

0xda157:
> Excellent, thank you! I've amended the commit to optimise the SVG.
> 
> The new guides include several methods for installing with Nix. Would it make more sense to have these?
> 
>     * Nix shell
> 
>     * Nix home manager
> 
>     * NixOS configuration

Sure, that would make more sense. Although I'd guess home-manager users would be able to figure it out given the Nix shell instructions, so I'll omit that.

> There seems to be 2 commands for Nix shell, and it looks like you're supposed to run both. Is that right?

No. One of them is Nix v2 cli and the other is the Nix v3 cli. I'll make that clear.

lpil:
Can we have just the v3 please 🙏

0xda157:
> Can we have just the v3 please 🙏

It's a bit of a complicated situation. The v3 cli is still experimental, but widely adopted and is preferred by much of the community.

lpil:
It came out 5 years ago and is very widely used, I think we can ignore the "experimental" tag the same as we ignore mature software using a `v0` version number. 😁

0xda157:
> It came out 5 years ago and is very widely used, I think we can ignore the "experimental" tag the same as we ignore mature software using a `v0` version number. 😁

I agree that the fact it's called experimental doesn't mean we shouldn't use it, but if we don't include the v2 cli version it will seem like enabling nix-command and flakes is required to add gleam to the PATH, which is not true.

lpil:
So could the v3 one be called "Nix flake" and the v2 be called "Nix somethingelse"?

src/website/install.gleam
```diff
@@ -462,7 +462,49 @@ Installing with asdf can take a long time as it builds Erlang from source. On ma
...
+      systems: [Linux, MacOs],
+      priority: MediumPriority,
+      content: "Gleam can be temporary added to the `$PATH` with `nix-shell`:
+
+```
+nix-shell -p gleam beamPackages.erlang beamPackages.rebar3
+```
+
+or `nix shell`, if you have `nix-command` and `flakes` experimental features
+enabled:
```
0xda157: > So could the v3 one be called "Nix flake" and the v2 be called "Nix somethingelse"?

we could do something like this, although I prefer the current version.

````suggestion
      content: "Gleam can be temporary added to the `$PATH` with `nix-shell` (nix channels):

```
nix-shell -p gleam beamPackages.erlang beamPackages.rebar3
```

or using the flake cli, with `nix shell`:
````

lpil: Installation methods shouldn't have any conditional logic and each should be one method rather than multiple related ones. It's too easy for people to make mistakes with conditional logic, so all the decision making here comes beforehand when you are picking and OS and a method.

0xda157: Oh that makes a lot more sense, I've split it into two.

lpil (approved):
Beautiful! Thank you!!

PR #662, opened 2026-06-30
strip html tags out of the .. well username

cmnstmntmn:
fix a problem on this page here: https://gleam.run/news/single-file-gleam-beam-programs-with-escript/

Screenshot 2026-06-30 at 22 53 37


lpil (commented):
Hello! Thank you, but that is her username! It would be poor manners to edit the username of a supporter.

cmnstmntmn:
100% strongly agree with you, 
100% my eyes hurts, seeing those markdown/html unrendered tags 🥹

lpil:
Could be worse, we could have failed to escape them 😁

PR #668, opened 2026-07-15
Adjust grammar

spenserblack:
Just a grammar issue I stumbled on when reading your news.

- A double "have" from "we've haven't" expanding to "we have have not"
- Changed "haven't able" to "haven't *been* able"

This could probably be just 1 commit (I hit "commit" too soon on the web interface), so please squash :slightly_smiling_face:


ankddev (approved):

lpil (approved):
Thank you

PR #672, opened 2026-07-28
fix: allow nix codeblocks in djot content

khalidbelk:
Hey : ) ! While fixing something here I got this panic error when launching the project, which apparently is what's making the CI fail as well (introduced by [d87d300](https://github.com/gleam-lang/website/commit/d87d30029b8ddf90c1a573413ea0253d3ec0890b)). This should fix it !

image

image


lpil (approved):
Thank you!!

PR #673, opened 2026-07-28
fix(case-studies): swap inverted headers for strand and uncover articles

khalidbelk:
While navigating through the website and reading [here](https://gleam.run/case-studies/), I noticed the articles titles / subtitles were inverted.
This restores the headers back to their respective articles.

### DIFF
--- a/pages/case-studies/strand.djot
+++ b/pages/case-studies/strand.djot
@@ -1,7 +1,7 @@
 ---
-title = "No room for error"
-subtitle = "A case study of Gleam in production at Uncover"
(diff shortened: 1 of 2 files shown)


lpil (commented):
Thank you. Could you correct the meta-title also please

khalidbelk:
Your welcome ! Oh since the part right after the `|` was correct, I thought the whole line was too 🥲. It's fixed now : )

lpil (approved):
Thanks again!!

PR #679, opened 2026-08-14
add a guides index page

graphiteisaac:
Howdy howdy

Finally coming full circle on the guides and implementing the guides index page to help people find documentation more easily. I've put the current 3 guides in there with some interpretations of the titles, and tags which are certainly incomplete, but it gives us a good point to launch from.

image


lpil:
Target the branch with the guides plz 💜

PR #683, opened 2026-09-01
syntax-highlighting for the cheatsheet pages

yoshi-monster:
Highlight Gleam, Erlang, Elixir, Rust (by using Gleam) and Python (by using Elixir) code on their respective cheatsheet pages.


lpil (approved):
Thanks!!

PR #687, opened 2026-09-14
fix: og:url shouldn't have duplicated path slashes

### LINKED ISSUE #684: Opengraph URLs have double slash
I noticed there's a double `/` on the `og:url` property when looking at one of the blog posts. Seems to effect other routes I checked too, but not sure if it's all of them.

Image

ghostdevv:
Added a normalisation to the constructed url before it's used in `og:url`. It looks like all the `PathMeta` instances started with a `/` anyway, but I added a guard to be safe

Closes #684


lpil (approved):
Thank you!

